PR #6, opened 2020-02-28
Feature/templates changes

drobnikj:


drobnikj:
@jancurn It was there already in all templates

jancurn:
Oh right, my bad

PR #7, opened 2020-03-01
add Typescript "hello world"

pocesar:
Not much, but it's a start


jancurn:
Nice!

templates/hello_world_ts/package.json
```diff
@@ -0,0 +1,21 @@
...
+    },
+    "devDependencies": {
+        "@types/node": "^13.7.7",
+        "@types/cheerio": "^0.22.16",
+        "@types/puppeteer": "^2.0.1",
+        "typescript": "^3.8.3"
+    },
+    "scripts": {
+        "build": "tsc -p tsconfig.json",
+        "start": "node lib/main.js",
```
drobnikj: I guess you need to update npm start command because you need to build files the first.

templates/hello_world_ts/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+# Note that you can use any other image from Docker Hub.
+FROM apify/actor-node-basic
+
+# Second, copy just package.json and package-lock.json since they are the only files
+# that affect NPM install in the next step
+COPY package*.json ./
+
+# Install NPM packages, skip optional and development dependencies to keep the
+# image small. Avoid logging too much and print the dependency tree for debugging
+RUN npm --quiet set progress=false \
```
drobnikj: I think you need to build the code before you run it and it needs to be done during docker build.

templates/hello_world_ts/package.json
pocesar: yeah, usually you need to build locally (doing `npm run build` before pushing), so you don't add devDependencies on the image. the `tsconfig.json` generates the files to the `lib` folder

templates/hello_world_ts/Dockerfile
pocesar: if you think it's better to add typescript as a dependency instead of a devDependency, I'll do it. but since you can't test locally without building it first, it makes sense to build then push/commit

pocesar: if the apify-cli supports using ts-node as the runtime instead of node, so that would work (didn't test to see)

pocesar:
@drobnikj it seems that it's working. the drawback is that TS (and ts-node) need to be a `dependency` instead of a `devDependency`, but shouldn't be much of a hit, size-wise

mnmkng:
Actors are not packages, so they're typically installed with devDeps locally, but they are only installed with deps on the platform (because of production install flag in Dockerfile).

On the platform, I think that the transpilation should be done in the build process. That's what it's for. It doesn't matter for this template, but for a large project, it would be slow to first transpile everything and only then run it.

@pocesar Can we utilize the Dockerfile for platform build and `devDeps` for local "build", so that the template will always work out of the box? The solution you proposed works, but it feels hacky. Especially the `"npm start": "node -r ..."` command and just having TS as a dependency, which I think would be weird for TS users to see.

B4nan:
Superseeded by #84

PR #27, opened 2020-05-27
Feature/disable npm runtime error

metalwarrior665:
Adding extra `env` to each apify.json. 

The Dockerfiles CMDs fallback to parent image CMD so we don't need to update it here. I made also PR to our main Dockerfiles.

### DIFF
--- a/templates/example_basic_crawler/apify.json
+++ b/templates/example_basic_crawler/apify.json
@@ -1,1 +1,3 @@
-{}
+{
+    "env": { "npm_config_loglevel": "silent" }
+}
--- a/templates/example_hello_world/apify.json
+++ b/templates/example_hello_world/apify.json
@@ -1,1 +1,3 @@
-{}
+{
+    "env": { "npm_config_loglevel": "silent" }
+}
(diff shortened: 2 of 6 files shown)


pocesar:
locally, it will work just as it would on the platform. since the local SDK doesn't use Docker images, I think that's fine

jancurn:
@pocesar oh right, fair enough

mnmkng:
So the point of this is to get rid of the NPM error on local? I don't know, it's weird. First of all, it will only work when using `apify-cli` and second, it injects some env vars into the system from a very unexpected place.

metalwarrior665:
Yes, that's the point. To save terminal space and also the error is confusing for Node beginners.

CLI is injecting many env vars too.

metalwarrior665:
But if you think it would be better to do it from another place (directly from the CLI), then I'm all for it. It is a cosmetic thing ofc, nothing too serious.

mnmkng:
You know what, I'll just merge it because it's better than the current situation and we can figure out a better way later.

PR #76, opened 2021-06-14
feat: Add example Python actor template

fnesveda:
This PR adds a template with a simple example Python actor based on an upcoming Python so that users wishing to write actors in Python have something to start from. It probably won't run with the Apify CLI locally, because the Python Client doesn't support saving storage locally, but we'll soon add a feature to the Apify app which will enable prefilling the multifile editor from these templates, so it will be useful there.

This needs the `apify/actor-python` to be published, and for this we need to merge [this PR](https://github.com/apify/apify-actor-docker/pull/57) first, so I'm marking this as a draft for now, and I'll test this and then mark it as ready for review when that is merged.


templates/example_python/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# Print the installed Python version, pip version
+# and all installed packages with their versions for debugging
+RUN echo "Python version:" \
+ && python --version \
+ && echo "Pip version:" \
+ && pip --version \
+ && echo "Installing dependencies from requirements.txt:" \
+ && pip install -r requirements.txt \
+ && echo "All installed Python packages:" \
+ && pip --version
```
mnmkng, line 21: I assume this should be something else than `pip --version` no?

fnesveda: Yeah, good catch, forgot to update this line after copy paste.

mnmkng (approved):

mhamas (approved):

PR #80, opened 2021-08-30
npm 7 requires --all for list

metalwarrior665:
It will work how it used to work with npm 6
https://docs.npmjs.com/cli/v7/commands/npm-ls#all


mnmkng:
Does it work with NPM6 as well?

metalwarrior665:
> Does it work with NPM6 as well?

@mnmkng Yup, with NPM 6 `npm list` == `npm list --all`.

mnmkng (approved):

PR #83, opened 2021-11-01
chore: upgrade examples to docusaurus v2

vladfrangu:
We should test this with the new docs site before merging 👍


B4nan:
I guess we can close this one, as the examples are now included in the crawlee repository directly, we can adjust those later if needed (or even reopen/merge this if it makes sense, but we dont need to care about the docusaurus adjustments).

PR #84, opened 2021-11-08
Feature hello world ts

petrpatek:
Created typescript template with the multistage build. Inspired by #7 and internal typescript projects.


pocesar:
hey @petrpatek that's nice to have TS as first-class citizen in templates ❤️  
there are a couple of things with the current approach: 
* there are no source maps and the source itself is discarded, making hard to debug just through logs, since it will show the error line being `/dist/src/something.js:73:11`, since most of the code usually goes into `src` and `main.js` being the entrypoint of the actor. 
* the behavior of starting an actor should be the same, locally and inside the container. 

using the start script as `node -r ts-node/register/transpile-only src/main.ts` solves both of those issues, ts-node adds the sourcemap support out-of-the-box, but adds another 2 issues: bundle size and extra dependencies, and typescript dependency itself is 60MB, an extra second downloading through NPM. IMHO, for the SDK being a runtime, it's worth this dependency hassle. 😄 

on the other hand, the side-by-side sourcemaps can be generated using tsconfig.json settings, but every change requires rebuilding, it's very easy to forget doing this when not using the watch mode, and as actors can't have a "hot module replacement" like React or Angular, `tsc -w` isn't really suitable

templates/example_hello_world_ts/.eslintrc
```diff
@@ -0,0 +1,3 @@
+{
+    "extends": "@apify"
```
B4nan, line 2: this should use the TS eslint config, this is the JS one, isn't it?

B4nan (approved):
> there are no source maps and the source itself is discarded, making hard to debug just through logs, since it will show the error line being /dist/src/something.js:73:11, since most of the code usually goes into src and main.js being the entrypoint of the actor.

Not sure why you say there are no source maps. This uses the shared apify tsconfig and it has source maps enabled. 

https://github.com/apify/apify-tsconfig/blob/main/tsconfig.json#L29

Also the compiled code should go just to `dist`, not to `dist/src`, and I believe it already works this way. Btw entry point is not `main.js`, but `dist/main.js`. 

Let's not invent squared wheels, we should not publish nor run TS code in production. 

> on the other hand, the side-by-side sourcemaps can be generated using tsconfig.json settings, but every change requires rebuilding, it's very easy to forget doing this when not using the watch mode, and as actors can't have a "hot module replacement" like React or Angular, tsc -w isn't really suitable

In development ts-node should be used, instead of rebuilding the app without any benefit. Even if we could make watch mode/hmr work, I can't see any benefit in that for use, we are talking about scripts, not webapps, changing their code on runtime without fully restarting seems like a very bad idea.

> the behavior of starting an actor should be the same, locally and inside the container.

If you are referring to using `tsc && node` vs `ts-node`, I really can't agree with this. I've been using this approach happily for years, never had issue with it. You don't want to deal with (re)building everything locally during development, neither you want to deal with source maps or wrong links in stack traces - that is why you want `ts-node` as it will deal with all of it, as you get stack traces poiting to the TS code, and you don't need rebuilding.

This looks pretty much as my TIM actor setup, and I was more than happy with how it worked locally and in production. So for me this is a green light.

templates/example_hello_world_ts/.eslintrc
petrpatek: Oh yeah, my bad missed that one. Fixing it now.

templates/example_hello_world_ts/tsconfig.eslint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "./tsconfig.json",
```
B4nan, line 2: i guess we can use the root tsconfig directly, or not? the dedicated eslint tsconfig is usually there to allow checking test files too, here it does not add any benefits

petrpatek: Yes, it is redundant to specify it. I have Improved the .eslintr file to use minimal working config.

petrpatek:
@B4nan Can I merge it like this?

B4nan:
Let's see what others think, it looks good to me, and we can iterate on that before sdk v3 gets out.

templates/example_typescript/package.json
```diff
@@ -0,0 +1,28 @@
...
+        "eslint": "^7.0.0",
+        "ts-node": "^10.3.0",
+        "typescript": "^4.4.4"
+    },
+    "scripts": {
+        "start": "npm run start:dev",
+        "start:prod": "node dist/main.js",
+        "start:dev": "ts-node -T src/main.ts",
+        "build": "tsc",
+        "lint": "./node_modules/.bin/eslint ./src --ext .js,.ts",
```
vladfrangu, line 22: No need for the `./node_modules` path (which is also incompatible with Windows devs that want to run it iirc)

```suggestion
        "lint": "eslint ./src --ext .js,.ts",
```

templates/example_typescript/package.json
```diff
@@ -0,0 +1,28 @@
...
+        "lint:fix": "./node_modules/.bin/eslint ./src --ext .js,.ts --fix",
```
vladfrangu, line 23:
```suggestion
        "lint:fix": "eslint ./src --ext .js,.ts --fix",
```

templates/manifest.json
```diff
@@ -71,6 +71,16 @@
         "timeoutSecs": 0,
         "memoryMbytes": 256
       }
+    },
+    {
+      "name": "example_typescript",
+      "description": "Example: Hello world in Typescript - A simple Typescript actor from which you can start developing",
```
vladfrangu, line 77:
```suggestion
      "description": "Example: Hello world in TypeScript - A simple TypeScript actor from which you can start developing",
```

vladfrangu (changes requested):
Small nits, otherwise this looks like an awesome example for a TS actor 👍

petrpatek:
@vladfrangu Thanks for catching this!

vladfrangu (approved):

PR #90, opened 2022-04-25
Add getting started templates

mnmkng:
Adding templates that should be used in developer onboarding flow of the Console. It should work approximately like this:

1. Users are prompted to create an actor.
2. They are presented with 3 language choices - Node, TS, Python
3. Depending on choice, the proper getting started template is selected
4. They are prompted to complete the missing piece of code
5. Build
6. Prompted to go to Input - which will have input schema
7. Fill in input
8. Run
9. Output schema should kick in and show nice table
10. Prompt to either deploy code from GitHub (go to GitHub guide), from local (CLI guide) or continue browsing Console.

@B4nan pls check the TS. I skipped build, because the code would not build, as there is a piece missing.
@fnesveda pls check the Python. I used your original example and quite naively updated it to match the other templates.

### DIFF
--- a/templates/getting_started_typescript/Dockerfile
+++ b/templates/getting_started_typescript/Dockerfile
@@ -1,0 +1,33 @@
+# First, specify the base Docker image. You can read more about
+# the available images at https://sdk.apify.com/docs/guides/docker-images
+# You can also use any other image from Docker Hub.
+FROM apify/actor-node:16
+
+# Second, copy just package.json and package-lock.json since it should be
+# the only file that affects "npm install" in the next step, to speed up the build
+COPY package*.json ./
+
+# Install NPM packages, skip optional and development dependencies to
+# keep the image small. Avoid logging too much and print the dependency
+# tree for debugging
+RUN npm --quiet set progress=false \
+ && npm install --only=prod --no-optional \
+ && echo "Installed NPM packages:" \
+ && (npm list --only=prod --no-optional --all || true) \
+ && echo "Node.js version:" \
+ && node --version \
+ && echo "NPM version:" \
+ && npm --version
+
+# Next, copy the remaining files and directories with the source code.
+# Since we do this after NPM install, quick build will be really fast
+# for most source file changes.
+COPY . ./
+
+# Optionally, specify how to launch the source code of your actor.
+# By default, Apify's base Docker images define the CMD instruction
+# that runs the Node.js source code using the command specified
+# in the "scripts.start" section of the package.json file.
+# In short, the instruction looks something like this:
+#
+# CMD npm start
--- a/templates/getting_started_typescript/Dockerfile
+++ b/templates/getting_started_typescript/Dockerfile
@@ -1,33 +1,28 @@
-# First, specify the base Docker image. You can read more about
-# the available images at https://sdk.apify.com/docs/guides/docker-images
-# You can also use any other image from Docker Hub.
-FROM apify/actor-node:16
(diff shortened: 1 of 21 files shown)


templates/getting_started_typescript/Dockerfile
```diff
@@ -0,0 +1,33 @@
+# First, specify the base Docker image. You can read more about
+# the available images at https://sdk.apify.com/docs/guides/docker-images
+# You can also use any other image from Docker Hub.
+FROM apify/actor-node:16
```
B4nan, line 4: the TS dockerfile should use multistage build, compile code and run the `npm start:prod` explicitly, we already have that in the TS template, so lets just use the dockerfile from there

https://github.com/apify/actor-templates/blob/master/templates/example_typescript/Dockerfile

templates/getting_started_typescript/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "name": "getting-started-typescript",
+    "version": "0.0.1",
+    "description": "This is an example of an Apify actor.",
+    "dependencies": {
+        "apify": "^2.3.0"
+    },
+    "devDependencies": {
+        "@apify/tsconfig": "^0.1.0",
+        "ts-node": "^10.3.0",
+        "typescript": "^4.4.4"
```
B4nan, line 11: let's bump to latest
```suggestion
        "typescript": "^4.6.3"
```

templates/getting_started_typescript/src/main.ts
```diff
@@ -0,0 +1,33 @@
...
+    console.log('Loading input')
+    // Structure of input is defined in INPUT_SCHEMA.json.
+    const input = await Apify.getInput() as InputSchema;
+    console.log('First number: ', input.firstNumber);
+    console.log('Second number: ', input.secondNumber);
+
+    // 👉 Complete the code so that result is
+    // the sum of firstNumber and secondNumber.
+    // 👇👇👇👇👇👇👇👇👇👇
+    const result = ;
```
B4nan, line 22: what about keeping the code valid by setting the default to `0`?
```suggestion
    const result = 0;
```

test/templates.test.js
```diff
@@ -36,14 +34,15 @@ const checkTemplateStructureAndRun = async (actorName) => {
...
 
+const checkTemplateRun = async (actorName) => {
     process.chdir(actorName);
     spawnSync('apify', ['run']);
     process.chdir('../');
-
     // Check if actor run without errors
     expect(console.log.args.map((arg) => arg[0])).not.toContain('Error:');
-};
+}
```
B4nan, line 45: nit (i guess we dont lint the tests?)
```suggestion
};
```

B4nan (changes requested):
(checked just the TS one)

templates/getting_started_typescript/src/main.ts
mnmkng: Dunno, I wanted the run to fail if they don't fill in something, instead of showing wrong sum in the output.

templates/getting_started_typescript/Dockerfile
mnmkng: Yeah, I know, but for that, the file needs to compile, which it doesn't with the empty result. It should run fine with ts-node, no?

itstrangvu (approved):
Design-wise I approve this. As long as there is input and output schema, we're good. Code-wise cannot tell though.

B4nan (approved):

PR #92, opened 2022-05-23
feat: add crawlee ts templates

B4nan:
Adds 3 new TS templates for crawlee.

### DIFF
--- a/templates/crawlee_cheerio_typescript/.gitignore
+++ b/templates/crawlee_cheerio_typescript/.gitignore
@@ -1,0 +1,7 @@
+# This file tells Git which files shouldn't be added to source control
+
+.idea
+dist
+node_modules
+apify_storage
+memory_storage
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@apify/actor-templates",
-  "version": "0.1.3",
+  "version": "0.1.4",
   "description": "A manifest of Apify actor templates.",
   "main": "src/index.js",
   "files": [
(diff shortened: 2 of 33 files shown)


templates/crawlee_cheerio_typescript/.gitignore
```diff
@@ -0,0 +1,7 @@
+# This file tells Git which files shouldn't be added to source control
+
+.idea
+dist
+node_modules
+apify_storage
+memory_storage
```
mnmkng, line 7: Why does memory storage need a folder? 😄

mnmkng (approved):

templates/crawlee_cheerio_typescript/.gitignore
B4nan: So users can read the values? 

Under the hood it works similarly to the storage-local, instead of sqlite we store things in memory, but we also dump them to file system (via worker thread, so it does not affect performance much).

B4nan:
FYI the plan is to merge few more things to apify-ts during this week (mainly backports of fixes from v2 + `ctx.sendRequest()` helper) and change the versions to `beta` (keeping `next` dist tag for v3). In the meantime I will prepare the other two templates for browsers and on monday we should be ready for internal testing. Given the E2E tests we have right now I don't expect much friction, it should mainly help us to better understand the migration process.

Btw do I get it right that we will need to publish new CLI version for the templates to show up?

mnmkng:
AFAIK we don't need a new CLI version. It should automatically load from the manifest file that's on the templates GitHub.

PR #98, opened 2022-07-21
use npm ci instead of npm install

PatrikTrefil:
I suggest using `npm ci` in templates for faster and deterministic builds. All flags used (`--no-optional`, `--include=`, `--only`) have the same behavior for `npm ci` based on my testing. Sadly, they are not properly documented in the npm docs...


PatrikTrefil:
Npm tried to fix their docs based on my [issue](https://github.com/npm/cli/issues/5204), but the docs still don't include the options we are using (but they were not documented for the `npm install` command either).

PR #101, opened 2022-08-01
fix: getting started typescript template

drobnikj:
Fixed getting started typescript template, which cannot be build on platform because typecheck failed.
Snímek obrazovky 2022-08-01 v 17 34 54


B4nan (approved):

drobnikj:
I just add issue to improve tests to prevent this error see #102

PR #104, opened 2022-08-16
Fix: Fixed name of the crawler.

petrpatek:


B4nan:
those templates will be removed after #103 is merged

PR #108, opened 2022-10-13
Updates templates based on new output documentation

gippy:
Resolves this issues: https://github.com/apify/apify-core/issues/8237

Updated the templates to follow the new output schema definition with docs here: https://docs.apify.com/actors/development/output-schema#viewtransformation-object-definition

Tested it and both worker and UI already have it implemented.

### DIFF
--- a/templates/getting_started_node/.actor/actor.json
+++ b/templates/getting_started_node/.actor/actor.json
@@ -20,23 +20,20 @@
                     },
                     "display": {
                         "component": "table",
-                        "columns": [
-                            {
+                        "properties": {
+                            "sum": {
                                 "label": "Sum",
-                                "format": "number",
-                                "field": "sum"
+                                "format": "number"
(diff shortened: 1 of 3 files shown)


templates/getting_started_node/.actor/actor.json
```diff
@@ -20,23 +20,20 @@
...
                             },
-                            {
+                            "secondNumber": {
                                 "label": "Second number",
-                                "format": "number",
-                                "field": "secondNumber"
+                                "format": "number"
                             }
-                        ]
+                        }
```
mnmkng, line 36: Could we potentially leave this out completely? I'm not sure if I want to, it works as documentation, but I have a feeling we wanted to implement the label renaming automatically without the user having to split and capitalize camelCase.

mnmkng (approved):

PR #109, opened 2022-11-04
Fix output schema in getting started TS

novotnyj:
I was testing my changes on staging and noticed this:
Snímek obrazovky 2022-11-04 v 11 26 06

Turns out that the default output schema for `Getting started: TypeScript` is wrong - the display object contains another display object...


gippy:
Oh that is my fault, good catch

gippy (approved):

mnmkng (approved):

PR #110, opened 2022-11-09
feat: Use actor.json and .actor folder

novotnyj:
This PR moves Dockerfile, INPUT_SCHEMA.json, README.md into `.actor` folder. 
It also adds explicit declarations of readme, input and dockerfile into `.actor/actor.json` files.

closes https://github.com/apify/apify-core/issues/8602


mhamas (approved):

templates/getting_started_node/.actor/actor.json
```diff
@@ -3,7 +3,10 @@
     "name": "getting-started-actor",
     "title": "Getting Started Actor",
     "description": "Adds two integers.",
-    "version": "0.0.1",
+    "version": "0.0",
+    "input": "./INPUT_SCHEMA.json",
```
mnmkng, line 7: It's a nitpick, but could we use lowercase `input-schema.json` here pls? I've been waiting for this for eternity 😄

mhamas: We have hardwired this to actor as per https://github.com/apify/apify-worker/pull/661#discussion_r1012940871. 

https://github.com/apify/apify-worker/blob/develop/src/actor/act2_build_job.js#L75-L78

How important is this? We would have to add lowercase support in worker (easy) and (more importantly), add this handling to CLI.

I'm fine with only supporting `.actor/input_schema.json` have you @mnmkng , @mtrunkat and @jancurn reached consensus on this?

jancurn: IMHO it's better to be explicit, rather than implicit - so that if someone reads the actor.json, they will know input schema is related and means something. that's we made also README and Dockerfile explicit

jancurn: I think this makes sense, and we can use convention and examples to push people to use `input-schema.json` rather than `INPUT_SCHEMA.json`

mhamas: Ok, so we go with the system above. We will be releasing the new CLI that already works based on this tmrw.

mnmkng: Phew. I got a little scared here 😅

novotnyj: Renamed and changed references in `actor.json`

mnmkng (approved):

mnmkng:
Should we merge or not yet?

mhamas:
CLI has been release so this is good to merge I think.

PR #112, opened 2022-12-16
Install packages before copying dist folder

mvolfik:
This should __actually__ speed up cached builds on code-only changes

### DIFF
--- a/templates/getting_started_typescript/.actor/Dockerfile
+++ b/templates/getting_started_typescript/.actor/Dockerfile
@@ -21,9 +21,6 @@ RUN npm run build
 # Create final image
 FROM apify/actor-node:16
 
-# Copy only built JS files from builder image
-COPY --from=builder /usr/src/app/dist ./dist
-
 # Copy just package.json and package-lock.json
 # to speed up the build using Docker layer cache.
 COPY package*.json ./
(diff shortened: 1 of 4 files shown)


templates/getting_started_typescript/.actor/Dockerfile
```diff
@@ -41,6 +38,9 @@ RUN npm --quiet set progress=false \
     && npm --version \
     && rm -r ~/.npm
 
+# Copy built JS files from builder image
+COPY --from=builder /usr/src/app/dist ./dist
+
 # Next, copy the remaining files and directories with the source code.
```
B4nan, line 44: do we even need to second `COPY . ./`? i guess it will copy things like JSONs and files outside of `src` too, but it will include the `src` folder too? looks a bit weird

cc @vladfrangu

B4nan (approved):

templates/getting_started_typescript/.actor/Dockerfile
mvolfik: I can't think of any reason - README, input schema and similar things are taken directly from the code/repo, not the final image, right?

But I'd say it doesn't really hurt us much to keep it there, it's just a few text files, and saves users from surprises when they want to read some file in the actor, but it's suddenly not there after build and they might not be too familiar with Docker

vladfrangu: > do we even need to second `COPY . ./`?

We definitely need it for things like copying the storage directory, as well as any other files the user has externally, but maybe thats more of a task for the user to do?

mvolfik: > storage directory

In which case would you want that? And `storage` is in both `.dockerignore` and `.gitignore` anyways

PR #113, opened 2023-01-03
Update `python-start` template to use the Python SDK

fnesveda:
This updates the "Getting Started: Python" template to use the Apify SDK, plus it changes the project format so that we can support dependency installation and actor running in the Apify CLI.

Don't merge it yet, it still uses the beta version of the SDK without support for local storage emulation. It's only here to discuss the Python actor format, plus to allow testing the CLI.


templates/getting_started_python/.gitignore
```diff
@@ -0,0 +1,8 @@
+# This file tells Git which files shouldn't be added to source control
```
jirimoravcik, line 1: Looks quite slim compared to https://github.com/github/gitignore/blob/main/Python.gitignore

templates/getting_started_python/.actor/Dockerfile
```diff
@@ -3,9 +3,10 @@
 # You can also use any other image from Docker Hub.
 FROM apify/actor-python:3.9
```
jirimoravcik, line 4: Any particular reason for 3.9?

jirimoravcik (approved):

templates/getting_started_python/.actor/Dockerfile
fnesveda: Not really. I was planning to update it to `3.11`, but we don't have `3.11` stable yet, I would increase it after we release the Python SDK and the accompanying Docker images with it.

templates/getting_started_python/.gitignore
fnesveda: Good point, I'll make it a bit fatter

fnesveda:
@jirimoravcik I've updated this a bit to have more standardized commands in the Makefile, but I think we can't make it truly universal unless we have the `scripts` section in `actor.json`, hopefully soon.

jirimoravcik (approved):

templates/getting_started_python/.actor/Dockerfile
```diff
@@ -3,9 +3,10 @@
...
 
-# Second, copy just requirements.txt into the actor image,
-# since it should be the only file that affects "pip install" in the next step,
+# Second, copy just requirements.txt and Makefile into the actor image,
```
mvolfik, line 6: no more makefile

templates/getting_started_python/.actor/Dockerfile
```diff
@@ -3,9 +3,10 @@
...
+# since they should be the only files that affect "make install" in the next step,
 # in order to speed up the build
+COPY Makefile ./
```
mvolfik, line 9: no more makefile

templates/getting_started_python/src/main.py
```diff
@@ -0,0 +1,33 @@
+import asyncio
```
mvolfik, line 1: technically the asyncio import is not needed here now

templates/getting_started_python/src/main.py
```diff
@@ -0,0 +1,33 @@
...
+from apify import Actor
+
+
+async def main():
+    async with Actor() as actor:
+        # Get the value of the actor input
+        actor_input = await actor.get_input()
+
+        # Structure of input is defined in INPUT_SCHEMA.json
+        first_number = getattr(actor_input, "first_number", None)
```
mvolfik, line 12:
```suggestion
        first_number = actor_input.get("first_number")
```

^ this is equal to `actor_input["number"]` with None as default. `getattr` is for `actor_input.property`, which isn't the same in Python.

the disadvantage is that it will throw an error if input is None - so maybe let's also add
```py
if type(actor_input) != dict:
    raise ValueError("Input must be a dictionary")
```

before this line? (or maybe even add a hint that the input should be in the storage/kv/default/input.json, unless that is handled by the SDK)

mvolfik (changes requested):
Nice, this looks much cleaner than the makefile, I'd say :D Just some issues with the example code..

templates/getting_started_python/.actor/Dockerfile
```diff
@@ -26,5 +27,5 @@ RUN echo "Python version:" \
 COPY . ./
 
 # Specify how to launch the source code of your actor.
-# By default, the main.py file is run
-CMD python3 main.py
+# By default, the "python -m src" command is run
```
jirimoravcik, line 30:
```suggestion
# By default, the "python3 -m src" command is run
```

jirimoravcik (approved):

templates/getting_started_python/src/main.py
fnesveda: This was done precisely so that it doesn't throw an error when input is None and does not fail on first run, before you create the storages and so on. I'd like to have some default input in the templates, at least for these "getting started" ones, but I think it was removed a few months back, I need to ask around why.

mvolfik: Sorry, I didn't write it well, but what I wanted to say that this won't work when the input is provided - the loaded json is a dictionary, but getattr() (= `x.prop`) doesn't access dictionary fields (`x["field"]`)

fnesveda: Riiiight, I need to read things better. 🤦

Now I wonder - should `Actor.get_input()` always return something? If there is no `INPUT` key in the default key-value store, return an empty dict? It would make it a lot easier to work with.

mvolfik: I don't think we should fallback to empty dict.

In JS SDK, we return undefined when there's no input, so by following the same logic, in Python it can return None. OTOH, Python and JS differ in what is considered "idiomatic" - in Python, many things (such as accessing fields of dictionary with `x['field']`) raise an error, that you can catch. While in JS, exceptions often only used for really critical issues, and when something just doesn't exist, you get `undefined`. So by following this, I would even support throwing an exception directly in `get_input` when there isn't any.

fnesveda: I think it's actually valid that a run is started with no input, and so it should not throw an exception.

But I agree, let's keep returning `None` in that case, and then just do `input = Actor.get_input() or {}` in the template, so that it's explicit, but easier to work with down the road.

mnmkng: > I'd like to have some default input in the templates, at least for these "getting started" ones, but I think it was removed a few months back, I need to ask around why.

@fnesveda Because the platform does not respect input in local KVS and locally the actor does not respect input schema prefills and defaults. So it was just confusing people and we decided to get rid of the local inputs. But we need to find a better way to do this. Perhaps update CLI. Or have SDK itself check and validate input schema when running locally.

fnesveda: Maybe the CLI could check if there is any input schema, and if there are any defaults in it, and if there are, generate a `INPUT` key in the emulated default key-value store with those defaults. I think it would make the templates much easier to work with for beginners.

We have to have a function for generating the default input from input schema ready somewhere, the platform does it too.

mvolfik: > get rid of local inputs

@mnmkng sorry, what does that mean?

@fnesveda automatically providing input schema default if there's not input seems like too a bit too much magic to me. But it might be actually good if it wrote this default into the storage, so the next time they are there and user can edit it.

mnmkng: @mvolfik most templates used to have `/storage/key-value-stores/default/INPUT.json` file and some default input in it. But it was confusing users and we had a few related bugs in CLI, so we removed the default inputs from most templates. But maybe there still are some, I don't know and don't wanna check all of them 😅

drobnikj (approved):
Looks better than version with `.venv` 👍

fnesveda:
Thanks for the reviews everyone! I decided to split this into two PRs, [one](https://github.com/apify/actor-templates/pull/118) where we update the Python actor format (which can be merged immediately) and this one, where we just update the template to use the Python SDK instead of the API client (which will be merged after the SDK is released).

This is so that we can merge the Apify CLI support ahead of time, and we can use it for internal testing of the Python SDK easier.

PR #117, opened 2023-01-19
fix: fixing dockefile typos

mhamas:
`dockefile` --> `dockerfile`


novotnyj (approved):
My bad 🤦 Thanks for fixing it!

PR #118, opened 2023-01-23
New Python actor format

fnesveda:
This PR updates the Python actor format to the format we agreed to in https://github.com/apify/actor-templates/pull/113, but without updating the actual actor code to use the Python SDK (since it's not released yet). That will come in another PR.


fnesveda:
@jirimoravcik Could you please just sanity check this that there's nothing stupid here, so that I can merge it? I'd like to merge the CLI support later as well, so that we're ready for the SDK release.

jirimoravcik (approved):
Looks good to me

PR #120, opened 2023-01-26
Make template archives deterministic and stable when there are no source changes

fnesveda:
When you run `npm run build`, the template zips were changed even if there were no changes to the template source. This was super annoying and I had to deal with way too many merge conflicts because of this.

This fixes it by using the built-in `zip` command instead of `archiver`, which generates the same zip from the same source files. There might be some differences between the Linux and macOS versions, we'll see after the CI runs for the first time, so this might still not be perfect, but let's see.


mnmkng:
Hmm, interesting. I don't think I ever had to deal with merge conflicts on the archives. Probably because I never built them locally and always let the CI do it. But maybe that changed recently...

mnmkng (approved):

fnesveda:
Yeah, I added this `--template-archive-url` flag to `apify-cli`, to help with testing new templates in PR, so I usually build them locally and then push the zips to GitHub:

https://github.com/apify/apify-cli/blob/master/src/commands/create.js#L173-L177

And now there was a lot of activity in this repo, so there was a lot of conflicts when I was updating the Python templates.

mtrunkat (approved):

PR #124, opened 2023-02-16
Added `meta.templateId` to `actor.json` and `input_schema` naming unification

jbartadev:
- Added `actor.json` where missing to make sure we always have `meta.templateName` available for all actors created from our templates (for tracking purposes)
- `INPUT_SCHEMA.json`/`input-schema.json` -> `input_schema.json` naming unification


novotnyj (approved):

mnmkng (changes requested):
Pls don't rename `input-schema` back to `INPUT_SCHEMA`. It's an ugly thing that I hoped I killed already 😅 (The main reason, besides being ugly, is that it suggests that the file name is some sort of a constant that needs to be respected, but it doesn't. You can name it whatever you want.)

Regarding the `templateName`, do we really need it? Could we duck type it? What happens if we change the template under the name? Do we have to version the names? What if users change it in their actor?

The structure of `actor.json` is defined in the Actor philosophy, for what it's worth. We shouldn't randomly add stuff to it.

fnesveda:
Between `input-schema` and `input_schema`, we should have `input_schema`, since we try to have underscores in filenames whenever possible: https://www.notion.so/apify/Coding-standards-editor-setup-58912bbe2c0c4ad9961ae818179bcbd9#9f86cbe9e80d46e197414ca1ddba1046

mnmkng:
> Between input-schema and input_schema, we should have input_schema

👍 I don't have a preference, but note that this is userland and AFAIK it's way more common to use dashes than underscores in the JS world, but Python devs probably use underscores more? So it's gonna be quite random. That's why I think it's important the file names don't look like "constants".

mtrunkat:
This is the final version and we agreed to have snake case lowercase: https://github.com/apify/actor-specs/blob/master/pages/ACTOR.md#actor-file

As for the `templateName`:"
- We need to track what templates people use locally
- There are more options:
    - We could bundle in Mixpanel client and send statistics directly there. But it seems kind of shady to be sending telemetry / analytics data from CLI tool to the internet. So as correct behaviour I see sending this along with command `apify push`.
    - We could store this info in some file like `~/.apify/stats.json` and have there mapping of actor directories to initial templates and then submit this info along with `apify push`. This option is more invisible to the user and does not bring any complexity to them but not sure if benefit is so big.
    - But as the simplest solution is just to store this in `.actor/actor.json` and then submit it with `apify push`. Perhaps we could have there one field we could use for stuff like this called `stats` or `meta` or whatever else and have `templateName` there. But I am not really opinionated about this.
    - Maybe somebody sees another option?

fnesveda:
I like the `meta.templateName` approach 👍 Then we don't pollute the top-level object, it's more clear that this is just something "internal" and we don't have to have strict validation for it.

fnesveda (changes requested):
Looks good, can you please add a unit test that checks that the template names in `manifest.json`, in `actor.json`s and in the folder names are in sync?

valekjo:
@jbartadev Please ping @mtrunkatova when this is deployed.

fnesveda:
@jbartadev We merged the [change](https://github.com/apify/actor-templates/pull/126) to add template IDs, can you update this PR to use the IDs instead of the names?

fnesveda (approved):

PR #126, opened 2023-02-22
feat: add id to template manifest

mnmkng:
I propose to replace the current template `name` with `id`. In this PR, I'm only adding the IDs because by removing the names, I would break CLI for anyone who's using the `apify create --template=abc` flag. I would remove the `name` after some deprecation period.

Why is this useful?

1. The names don't follow any meaningful structure so you always have to open the manifest to find the template name. The IDs follow a pattern so they're a bit less random.
2. The name ID implies that changing them could break something. Yesterday I had a chat with @jbartadev and we figured that changing the `name` does not break anything, but today I learned that I would break the CLI flags for people.
3. Given the structure of the IDs, there's much less need to rename anything in the future, which should prevent future breaking changes.

In the CLI, I will add a new flag called `templateId` so the change won't be breaking there and I'll deprecate the `template`. This whole thing actually started with me not understanding what should go into `template` and I learned that `template` is actually `templateName`.


mnmkng:
Bah, sorry about the formatting changes in the same commit. Missed that this happened.

fnesveda (approved):
Please call the flag `template-id` and not `templateId`, to be consistent with the naming of the other flags we have.

fnesveda:
Also @jbartadev maybe we should put the IDs in the `meta` field instead of the name.

fnesveda:
Hm, and we should also rename the template folders by the ID instead of the name.

jbartadev:
> Hm, and we should also rename the template folders by the ID instead of the name.

In that case, we would need the ID at all no? We'd be back on matching the id with the folder name (which is the current state). Basically, we could just rename the template folder files and templates `name` in the `manifest.json` and the effect would be the same. Or am I missing something? @fnesveda

fnesveda:
I think that the desired final state is:
- don't have `name` in the manifest at all
- have just `id` in the manifest, so you can filter templates by ID in the CLI
- have the template folders named the same as the template `id`, for consistency

Is that right @mnmkng

mnmkng:
Yeah, that's right. The folder names should be irrelevant to external consumers, but it makes sense to match their names to the template ids.

jbartadev (approved):

mnmkng:
Made the requested changes. Just checking for the last time if I can merge it like this. With folder names and archive URLs renamed. I also added missing technologies and browser names to the IDs

jbartadev (approved):
LGTM

fnesveda (changes requested):
The changes themselves look good, but there will be problems with the deployment, since the URLs to the ZIPs are changing.

One problem is that the URLs to the ZIPs are changing, but the ZIPs are not changed in this PR, so until the CI builds the templates and commits them, the URLs in the manifest will point to non-existing templates, and if someone tries to create an actor in the CLI before the ZIPs are committed, it will fail for him. The fix here is easy, just run `npm run build` locally and commit the updated ZIPs to this PR.

The bigger problem is that the Console caches the template manifest indefinitely, and it caches the template ZIPs for 30 minutes. So once we merge this, if anyone will try to create an actor and the template ZIPs are not cached, downloading them will fail, because it will try to download them from the old URL. We should change the caching logic in the Console to prevent these issues in the future, but for now, we can:
1. rotate the Console Kubernetes deployment
2. wait ~20 minutes until most of the templates are cached on the Console pods
3. merge this PR
4. rotate the Kubernetes deployment again so that the new manifest and new templates are cached on the Console pods

Let's merge this on Monday, so that we can sync on this in person.

fnesveda (approved):

PR #128, opened 2023-03-03
Feature: Add a Python + Scrapy template

fnesveda:
This adds a Python + Scrapy template.

It uses the default Scrapy project structure, and is based on an example straight from Scrapy: https://github.com/scrapy/quotesbot

It's runnable both as an actor, and also directly via Scrapy:
`scrapy crawl quotes -a tag=humor -o quotes.json`

The Scrapy project can actually remain completely untouched, we can just add a few files around it to make it into an actor.

Once we agree on the code I'd probably publish it as an example actor too.

BTW, figuring out the logging setup was a pain in the ass.

Also, the Scrapy example is published with the MIT license, we should include it somewhere here in the repo, to be completely correct.


mtrunkat (approved):

templates/python-scrapy/src/__main__.py
```diff
@@ -0,0 +1,58 @@
...
+# We can't attach our log handler to the loggers normally,
+# because Scrapy would remove them in the `configure_logging` call here:
+# https://github.com/scrapy/scrapy/blob/a5c1ef82762c6c0910abea00c0a6249c40005e44/scrapy/utils/log.py#L95
+# (even though `disable_existing_loggers` is set to False :facepalm:).
+# We need to monkeypatch Scrapy's `configure_logging` method like this,
+# so that our handler is attached right after Scrapy calls the `configure_logging` method,
+# because otherwise we would lose some log messages.
+old_configure_logging = scrapy.utils.log.configure_logging
+
+def new_configure_logging(*args, **kwargs):
```
jirimoravcik, line 31: Nice

jirimoravcik (approved):

PR #130, opened 2023-03-08
feat: adding cypress template

katacek:
solving #9982 -> creating actor template for cypress testing

The actor itself can be tested [here](https://console.apify.com/actors/FN32HgKUjcZCYhK8y/runs/guPGyU5dgLGN0BA8H#output) or by creating new actor using [this](https://github.com/apify/actor-templates#feat/add-cypress-actor-template:templates/js-cypress) url.

Screenshot 2023-03-09 at 14 51 57


templates/js-cypress/.actor/actor.json
```diff
@@ -0,0 +1,16 @@
+{
```
valekjo, line 1: Broken formatting here.

templates/js-cypress/cypress/e2e/first-spec.cy.js
```diff
@@ -0,0 +1,7 @@
+describe('first tests suite', () => {
+    it('visit Apify main page', () => {
+        cy.visit('https://apify.com');
```
valekjo, line 3: I'd make the baseUrl part of the input and pass it via config, so you would only use relative links in visit.

templates/js-cypress/main.js
```diff
@@ -0,0 +1,55 @@
+import { Actor } from 'apify';
+import fs from 'fs';
+import cypress from 'cypress';
+import { globby } from 'globby';
+
+await Actor.init();
+
+// todo: be able to send everything in input
+const { video, defaultCommandTimeout, viewportHeight, viewportWidth, ...rest } = await Actor.getInput();
```
valekjo, line 9: Since we don't use the parameters elsewhere, we can pass the full input as config.

templates/js-cypress/main.js
```diff
@@ -0,0 +1,55 @@
...
+
+const runOneSpec = (spec) => {
+return cypress.run({
```
valekjo, line 12: Broken formatting here (and elsewhere), probably just eslint setup locally?

templates/js-cypress/main.js
```diff
@@ -0,0 +1,55 @@
...
+        viewportWidth,
+        ...rest,
+    },
+    spec,
+})
+}
+
+const tests = await globby('./cypress/e2e/*-spec.cy.js');
+
+console.log(`Running tests from following files: ${tests}`)
```
valekjo, line 26: It's better to use apify log instead (it's in other templates probably)

templates/js-cypress/main.js
```diff
@@ -0,0 +1,55 @@
...
+    const result = await runOneSpec(test);
+    let keyValueStoreLink = '';
+    if (result.config.video) {
+        const baseName = test.split('/').pop();
+        const file = `./cypress/videos/${baseName}.mp4`;
+        const kvsKeyName = baseName.replaceAll('.', '-');
+        await kvs.setValue(kvsKeyName, fs.readFileSync(file), { contentType: 'video/mp4' });
+
+        keyValueStoreLink = await kvs.getPublicUrl(kvsKeyName);
+        result.keyValueStoreLink = keyValueStoreLink;
```
valekjo, line 40: I wouldn't store the link on the result, but rather on transformed result, it would be nice to keep the raw cypress output as it is.

templates/js-cypress/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "name": "cypress-module-api",
+  "version": "1.0.0",
+  "type": "module",
+  "description": "Run Cypress via its module API",
+  "engines": {
+    "node": ">=16.0.0"
+  },
+  "dependencies": {
+    "cypress": "^12.7.0",
+    "console.table": "^0.10.0",
```
valekjo, line 11: This is not used in the end, so I'd drop it.

templates/js-cypress/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "cypress-module-api",
+  "version": "1.0.0",
+  "type": "module",
+  "description": "Run Cypress via its module API",
+  "engines": {
+    "node": ">=16.0.0"
+  },
+  "dependencies": {
+    "cypress": "^12.7.0",
```
valekjo, line 10: I'm not sure we need to include cypress if it's already in the base image, we should test this.

templates/js-cypress/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "console.table": "^0.10.0",
+    "globby": "^13.1.3",
+    "apify": "^3.0.0"
+  },
+  "devDependencies": {
+    "@apify/eslint-config": "^0.3.1",
+    "eslint": "^8.20.0"
+  },
+  "scripts": {
+    "cypress:open": "../../node_modules/.bin/cypress open",
```
valekjo, line 20: If we keep cypress in dependencies, just `cypress open` should be enough.

Or maybe we could make `cypress` dev dependency, because on prod it should use the comand that is already in cypress image.

templates/manifest.json
```diff
@@ -200,6 +200,22 @@
...
+            "label": "Cypress",
+            "category": "javascript",
+            "technologies": [
+                "nodejs"
+            ],
+            "description": "Empty starter actor. A blank slate with minimal starting code for developing actors from scratch using JavaScript.",
+            "archiveUrl": "https://github.com/apify/actor-templates/blob/feat/add-cypress-actor-template/dist/templates/js-cypress.zip?raw=true",
+            "defaultRunOptions": {
+                "build": "latest",
+                "memoryMbytes": 2048,
```
valekjo, line 216: Would change this to something higher that works fast enough for the tests to pass.

But probably should be in the free plan limits, so 4GB

templates/manifest.json
```diff
@@ -200,6 +200,22 @@
...
+        {
+            "id": "js-cypress",
+            "name": "project_cypress",
+            "label": "Cypress",
+            "category": "javascript",
+            "technologies": [
+                "nodejs"
+            ],
+            "description": "Empty starter actor. A blank slate with minimal starting code for developing actors from scratch using JavaScript.",
+            "archiveUrl": "https://github.com/apify/actor-templates/blob/feat/add-cypress-actor-template/dist/templates/js-cypress.zip?raw=true",
```
valekjo, line 213: This needs to be switched back to master branch, using other branch name is just for local testing.

valekjo (changes requested):
👏

templates/js-cypress/cypress/e2e/first-spec.cy.js
katacek: Hmm, just thinking about usecase when I would like to upload more tests with different `baseUrls` and run them at once? Then, I would need to separate them and create tasks and adjust the test itself.. 

But if you think that it is edge case and users usually go with one base url, then it makes sense and I would adjust it 👍

templates/js-cypress/.actor/actor.json
katacek: hmmm, true, eslint is apparently not working 🤔

templates/js-cypress/package.json
katacek: moving cypress to devDependencies resulted in following error (see the package.json prnt scrn and error prnt scrn) 🤔  moving it back to dependencies for now

Screenshot 2023-03-10 at 16 21 26

Screenshot 2023-03-10 at 16 22 58

templates/js-cypress/package.json
katacek: keeping cypress open, deleting `"start": "node ./main.js"`

katacek: actually keeping the start as well, otherwise, it cannot be run locally by `apify run`, I probably misunderstood it

templates/js-cypress/cypress/e2e/first-spec.cy.js
katacek: added to input schema as required field with pre-fill

templates/js-cypress/main.js
```diff
@@ -0,0 +1,49 @@
...
+        totalPending: result?.totalPending,
+        totalFailed: result?.totalFailed,
+        totalSkipped: result?.totalSkipped,
+        totalDuration: result?.totalDuration,
+        videoLink: keyValueStoreLink,
+        rawData: result,
+    };
+    await dataset.pushData(transformedResult);
+}
+
```
jbartadev, line 48: It'd be cool to also log the result of the tests here. Something like:

```
Results:

10 of 10 passed (100%)
 
```

jbartadev: Maybe even what tests passed/failed so you can see it in the log without coming to the kv store => better UX

templates/js-cypress/main.js
```diff
@@ -0,0 +1,49 @@
...
+import fs from 'fs';
+import cypress from 'cypress';
+import { globby } from 'globby';
+import log from '@apify/log';
+
+await Actor.init();
+
+const input = await Actor.getInput();
+
+const runOneSpec = (spec) => {
```
jbartadev, line 11: I'd prefer:

`const runOneSpec = (spec) => cypress.run({ config: input, spec });`

but up to you

templates/js-cypress/cypress/e2e/first-spec.cy.js
```diff
@@ -0,0 +1,7 @@
+describe('first tests suite', () => {
+    it('visit Apify main page', () => {
+        cy.visit('/');
+        cy.log('Visiting Apify main page');
+        cy.contains('h1', 'Build reliable web scrapers. Fast.').should('exist');
```
jbartadev, line 5: Wouldn't be better to use something more generic that won't break if we change the title on the web?

templates/js-cypress/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "cypress-module-api",
+  "version": "1.0.0",
+  "type": "module",
+  "description": "Run Cypress via its module API",
+  "engines": {
+    "node": ">=16.0.0"
```
jbartadev, line 7: You have `18.12.0` in the `.nvmrc` so maybe it could be `"node": ">=18.0.0" `?

templates/js-cypress/.actor/input_schema.json
```diff
@@ -0,0 +1,44 @@
...
+        },
+        "defaultCommandTimeout": {
+            "title": "Default Command Timeout",
+            "type": "integer",
+            "description": "Maximum time the for the command to be executed. Default is `30000`.",
+            "default": 30000,
+            "sectionCaption": "Advanced settings",
+            "sectionDescription": "These settings are not required for most use cases. Use them to customize or debug your test run."
+        },
+        "video": {
```
jbartadev, line 36: We could also add capturing screenshots when a failure in the future 👍

jbartadev (changes requested):
Nice! Good job there! Just a couple of small comments/suggestions/questions

valekjo (approved):

templates/js-cypress/main.js
katacek: Great idea, I have added the Summary table at the end of the log

Screenshot 2023-03-15 at 15 08 10

templates/js-cypress/cypress/e2e/first-spec.cy.js
katacek: sure, changed to 
`cy.contains('div', 'Apify').should('exist');`
I suppose that would be safer. Or any other ideas are welcome :-)

templates/js-cypress/package.json
katacek: goood catch 👌

templates/js-cypress/main.js
jbartadev: This is cool! 👍

templates/js-cypress/main.js
```diff
@@ -0,0 +1,62 @@
+import { Actor } from 'apify';
+import fs from 'fs';
+import cypress from 'cypress';
+import { globby } from 'globby';
+import log from '@apify/log';
+import 'console.table';
```
jbartadev, line 6: Just a question, do you really need to import this here?

templates/js-cypress/main.js
```diff
@@ -0,0 +1,62 @@
...
+
+await Actor.init();
+
+const input = await Actor.getInput();
+const runOneSpec = (spec) => cypress.run({ config: input, spec });
+
+log.info(`Running tests with following input: ${JSON.stringify(input)}`);
+
```
jbartadev, line 14: The spacing and logical continuity of this feels weird to me. I'd go for something like:

```
const runOneSpec = (spec) => cypress.run({ config: input, spec });

await Actor.init();

const input = await Actor.getInput();
log.info(`Running tests with following input: ${JSON.stringify(input)}`);

const tests = await globby('./cypress/e2e/*-spec.cy.js');
log.info(`Getting tests: ${tests}`);
...
```

What do you think?

templates/js-cypress/main.js
```diff
@@ -0,0 +1,62 @@
...
+import fs from 'fs';
+import cypress from 'cypress';
+import { globby } from 'globby';
+import log from '@apify/log';
+import 'console.table';
+
+await Actor.init();
+
+const input = await Actor.getInput();
+const runOneSpec = (spec) => cypress.run({ config: input, spec });
```
jbartadev, line 11: Also, maybe a comment on what exactly it does? This is a template, the more comments and explanations the better

jbartadev (changes requested):
Much cooler now! Just nitpicks and good to go!

templates/js-cypress/main.js
katacek: it seems yes, if not there, the summary table is not drawn

Screenshot 2023-03-16 at 13 16 42

jbartadev: Interesting 🤔

jbartadev (approved):
LGTM, good job! 😺

PR #133, opened 2023-03-13
feat: add Playwright Test Runner template

barjin:
Adds the Playwright Test Runner actor template.

Btw what is the use case here? The current actor is basically an Apify wrapper for Playwight Test Runner - accepts all the inputs through the input schema and outputs the results to dataset / KVS. Unlike Crawlee, it isn't really a "multi-use" framework to build your projects on - all the projects you can build out of this template look exactly the same to me 🤷🏼 

So... am I missing something, or is this just an SEO visibility stunt? 😄


metalwarrior665:
@barjin I think the answer is mostly yes :/ But maybe there is a way to generalize this, add some unique functionality?

B4nan:
cc @mtrunkat

B4nan:
To me, this is about developing a test suite locally and pushing the actor via CLI, which for a developer is imho easier than going through the UI in our console app. 

The template shouldn't accept the test as an input parameter - the point of it is to develop the tests locally, there should be out of box setup for more than a single test, so I guess a folder and an NPM script to run the test generator (configured to dump the result into that folder)?

templates/ts-playwright-test-runner/README.md
```diff
@@ -0,0 +1,51 @@
+# Playwright Test
+
+Run your Playwright tests on the Apify Platform effectively and easily. Just set up your test environment using a user-friendly UI and let the platform do the rest.
+
+> Note: This is a custom version of this actor. If you want to use the original simple version, please use the [official actor](https://apify.com/jindrich.bar/playwright-test).
+> 
+> Unlike the original actor, this version reads test suite files from the `tests` folder and does not allow you to pass the test codes via Apify input.
```
B4nan, line 7:
```suggestion
> Note: This is a custom version of [`Playwright Test Runner`](https://apify.com/jindrich.bar/playwright-test) actor. Unlike the original actor, this version reads test suite files from the `tests` folder and does not allow you to pass the test codes via Apify input.
```

B4nan (approved):
few nits

templates/ts-playwright-test-runner/.actor/actor.json
```diff
@@ -8,8 +8,10 @@
...
         "templateId": "ts-playwright-test-runner"
     },
-	"environmentVariables": {},
 	"storages": {
-        "dataset": "./OUTPUT_SCHEMA.json"
-    }
+        "dataset": "./output_schema.json"
+    },
+	"input": "./input_schema.json",
+    "readme": "../README.md",
```
B4nan, line 15: you can drop this line, we had a short discussion about this today and I've removed it from all the other templates too

PR #135, opened 2023-03-13
Post-create messages for Python actors

fnesveda:
Python actors require some post-create instructions:
- how to add additonal dependencies
- how to install Playwright and its browsers

This PR adds thse post-create messages with specific instructions to the Python actor templates.

The accompanying PR for Apify CLI is here: https://github.com/apify/apify-cli/pull/349


jirimoravcik (approved):

mtrunkat (approved):
I don't know what the message is talking about but looks legit :)

templates/manifest.json
```diff
@@ -6,6 +6,9 @@
             "label": "Start with Python",
             "category": "python",
             "description": "Project template using the Apify SDK for Python",
+            "messages": {
+                "postCreate": "To install additional Python packages, you need to activate the virtual environment in the \".venv\" folder in the actor directory."
```
drobnikj, line 10: What about adding a command on how to do it or a link with instruction or do you think the most developer know it?

drobnikj (approved):
Just idea about instruction for
`python3 -m venv env`

templates/manifest.json
fnesveda: Well the problem is it's platform-specific, it's different on Windows and on Linux, so I wanted to avoid some complicated logic and let the user do it the way they know.

fnesveda: Ideally we would show them the correct instructions for their OS and their shell, but I think that would be quite complex to do, maybe we can do it later.

PR #136, opened 2023-03-15
feat: add template replacements, update descriptions and labels

mnmkng:
This PR is mostly about unifying the template labels and improving descriptions, but I also thought it might be useful to add another field to the manifest: `replacements` (name up for discussion).

The purpose of replacements is to have CLI redirects between templates. Now if we remove a template, CLI integrations that use `-t` will break. With the replacements field, we could redirect them to new/updated/similar templates. I'm not sure if it's super important, but it feels a bit safer than simply deleting templates.

I removed the language from most labels, because it felt super repetitive and especially in the CLI it does not make sense, because you first select the language and then you would see ... Python, ... Python ... Python everywhere. Let me know what you think.

### DIFF
--- a/templates/manifest.json
+++ b/templates/manifest.json
@@ -244,7 +244,7 @@
         {
             "id": "js-empty",
             "name": "project_empty",
-            "label": "Start from scratch with JavaScript",
+            "label": "Start from scratch",
             "category": "javascript",
             "technologies": [
                 "nodejs"
--- a/templates/manifest.json
+++ b/templates/manifest.json
@@ -14,7 +14,10 @@
                 "build": "latest",
                 "memoryMbytes": 1024,
                 "timeoutSecs": 3600
-            }
+            },
+            "aliases": [
+                "getting_started_python"
+            ]
         },
         {
             "id": "python-beautifulsoup",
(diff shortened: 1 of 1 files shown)


templates/manifest.json
```diff
@@ -254,5 +257,12 @@
                 "timeoutSecs": 3600
             }
         }
+    ],
+    "replacements": [
+        {
+            "fromKey": "name",
+            "from": "getting_started_python",
+            "to": "python-start"
+        }
```
fnesveda, line 266: Maybe we could have this as `aliases` directly on the template objects? Then the CLI could just go through the templates and find one matching by ID or by alias.

fnesveda: Then we could add the `name` of each template as an alias, so that we don't break backwards compatibility.

fnesveda (changes requested):
Cool! One idea about where to put the replacements / aliases.

templates/manifest.json
mnmkng: Ha! I had a similar idea that the CLI should look at `id` when `name` isn't available (hence the same ids an names for all new templates), but this takes it even further and I like it. 👍

PR #139, opened 2023-03-28
feat(app): add showcase files to manifest.json

HonzaTuron:
Adding showcase files to show them in template details (as we don't want to show all of them). Design and arguments described in [following issue](https://github.com/apify/apify-web/issues/2545).

Feel free to suggest different/more files for showcase.

### DIFF
--- a/templates/manifest.json
+++ b/templates/manifest.json
@@ -114,11 +114,9 @@
                 "timeoutSecs": 3600
             },
             "showcaseFiles": [
-                "src/main.py",
                 "src/__main__.py",
-                "src/items.py",
+                "src/main.py",
                 "src/pipelines.py",
-                "src/settings.py",
                 "src/spiders/title_spider.py"
             ]
         },
--- a/templates/manifest.json
+++ b/templates/manifest.json
@@ -17,6 +17,10 @@
             },
             "aliases": [
                 "getting_started_python"
+            ],
+            "showcaseFiles": [
+                "src/main.py",
+                "src/__main__.py"
             ]
         },
         {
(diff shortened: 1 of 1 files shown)


templates/manifest.json
```diff
@@ -17,6 +17,10 @@
             },
             "aliases": [
                 "getting_started_python"
+            ],
+            "showcaseFiles": [
+                "src/main.py",
+                "src/__main__.py"
```
fnesveda, line 23: For the Python ones, I would not show the `src/__main__.py` file, the logic there isn't really that necessary to show off.

templates/manifest.json
```diff
@@ -96,7 +112,15 @@
...
                 "timeoutSecs": 3600
-            }
+            },
+            "showcaseFiles": [
+                "src/main.py",
+                "src/__main__.py",
+                "src/items.py",
+                "src/pipelines.py",
+                "src/settings.py",
+                "src/spiders/title_spider.py"
```
fnesveda, line 122: Here I would only show `src/main.py`, `src/pipelines.py` and `src/spiders/title_spider.py`, the rest are not that important.

fnesveda (changes requested):
Cool! I had some suggestions about which files to show for the Python actors.

templates/manifest.json
mnmkng: I guess we sort the files by the order in the array right? Even though I don't know much about Python, it would be weird for me to find that some files are seemingly missing from the template. Or, the other way around, that some files were added to my actor even though they were not in the list.

I kinda understand that for supporting files, like eslint.json or something like that. But I think we should always show all the source files. Would be weird to e.g. find an import of a file that's not visible in the template.

templates/manifest.json
mnmkng: same concern as above comment

HonzaTuron: In that case, i keep __main__.py as well to unify it with the rest of python scrapers.

jbartadev (approved):

fnesveda (approved):

PR #140, opened 2023-03-29
feat: add BeautifulSoup template README

PerVillalva:
I'm starting with the Python SDK templates and will progressively commit the content for the remaining templates to this PR.

Please, let me know if you have any feedback regarding the structure of the README.

The content of the README will be displayed on the new Template Detail page, planned by the nux-team. For more information, please refer to the [design and discussion in Figma](https://www.figma.com/file/rgPzqcsFZYQH6QgvUTsPYS/Template-detail?node-id=796-58149&t=q6XLq7jA69sMHPJ2-0).


mnmkng:
I wonder whether including all those commands in the readme's is the best way to go. When some of them change, we would have to change them in all the readmes. And I guess over time, there will be more "shared" logic.

- I suggest we either move this out from the readme and to web/platform. Something like the API tab in Actors. It's also auto-generated.
- Or, we could add another build script to this repo. The build script would take the template's readme without the commands and it would dynamically add all the shared parts. Then, in the built templates (zip files), the readmes would be full length.

fnesveda (changes requested):
Is it necessary to have the code of the template copied to the README? I thought the NUX team is solving it in a different way: https://github.com/apify/actor-templates/pull/139#pullrequestreview-1362911132

PerVillalva:
Good point @mnmkng. I'm following the current design, and the commands are the bulk of the information displayed on the Template Detail page. And, as far as I know, the content that will populate the Template Detail page is supposed to come from the template README.

I agree with your comment but I'm just not sure how I can go about solving this issue or with whom I should discuss it.

PerVillalva:
@fnesveda Thanks for the info, Franta. I removed the code from the README.

templates/readmeTemplate.mustache
```diff
@@ -0,0 +1,66 @@
...
+## Documentation reference
+
+---
+
+To learn more about Apify and Actors, take a look at the following resources:
+
+- [Apify SDK for {{sdkLanguage}} documentation](https://docs.apify.com/sdk/{{sdkDocsPath}})
+- [Apify Platform documentation](https://docs.apify.com/platform)
+- [Join our developer community on Discord](https://discord.com/invite/jyEM2PRvMU)
+{{/template}}
```
HonzaTuron, line 66: Please add newline here

HonzaTuron (approved):
To describe current solution - readme templates are generated once user installs packages through `npm install`. The rest of the logic remains same (need to be build with build script before commit and push)

fnesveda (changes requested):
I'm still not sure this makes sense to include these instructions in the templates that will be downloaded when you run `apify create`. If you already do `apify create`, then you don't need the instructions because you've already done them.

PerVillalva:
@fnesveda From my understanding, the primary reason for having these instructions on the README is that the README's content will be used to populate the content displayed on the dedicated template pages that will be added to the console ([see discussion in Figma](https://www.figma.com/file/rgPzqcsFZYQH6QgvUTsPYS/Template-detail?node-id=796-58149&t=f3QGvsQ3YBLQT3Nt-0)). And people who land on that page might not necessarily have run `apify create`.

fnesveda:
@PerVillalva That I understand, but then what's the point of rendering the READMEs here? Why can't we just load the templates manifest on the web, and render the template pages there directly? It just feels too complicated this way.

PerVillalva:
Hm, good point, @fnesveda. I was doing this way just because that's how I knew how to do it. 

@HonzaTuron do you have any insights on how we could load the templates manifest on the web, and render the template pages there directly?

fnesveda:
OK I guess, still feels unnecessary, I doubt that many people visit the actual templates source here.

HonzaTuron:
@PerVillalva We have markdown rendere set-up on web so it would be enough to use it.

PerVillalva:
@fnesveda Would you be ok with this solution for the time being? If necessary, we can revisit and improve it later.

fnesveda:
Sure, it's not really a big deal

package.json
```diff
@@ -8,7 +8,8 @@
   ],
   "scripts": {
     "test": "jest --testTimeout=60000 --forceExit",
-    "build": "node ./src/build.js"
+    "build": "node ./src/build.js",
+    "postinstall": "node ./templates/readmeTemplate.mjs"
```
fnesveda, line 12: This will fix the tests failing:
```suggestion
    "postinstall": "cd templates && node ./readmeTemplate.mjs"
```

fnesveda (changes requested):
One fix to the tests:

fnesveda (approved):

PR #141, opened 2023-04-18
Increase memory for Python templates that use browsers

fnesveda:
People running actors made from these templates had their runs super slow because running a browser with 1GB of memory is not a great experience. This should make it better for them.


jirimoravcik (approved):
good point

mtrunkat (approved):

PR #144, opened 2023-05-04
feat: Adding a js-langchain template draft

mtrunkat:


templates/js-langchain/.eslintrc
mnmkng, line 1: Just yesterday we agreed that we will remove ESLint from all but "bootstrap" templates, because users might have other opinions about ESLint rules and it slows down installs, builds and so on.

Bootstrap templates are templates that aren't really examples, but project skeletons that should help a power user start a new project faster.

templates/js-langchain/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "dependencies": {
+        "apify": "^3.0.0",
+        "hnswlib-node": "^1.4.2",
+        "langchain": "^0.0.82",
+        "tar": "^6.1.14"
+    },
+    "devDependencies": {
+        "@apify/eslint-config": "^0.3.1",
+        "eslint": "^8.36.0"
+    },
```
mnmkng, line 19:
```suggestion
```

mnmkng (changes requested):

templates/js-langchain/.eslintrc
mtrunkat: Makes sense

mtrunkat: Have you also discussed .editorconfig? It probably also makes sense to remove those.

mnmkng: Yeah, editorconfig makes sense to me as well.

templates/js-langchain/.actor/Dockerfile
```diff
@@ -0,0 +1,32 @@
+# Specify the base Docker image. You can read more about
+# the available images at https://docs.apify.com/sdk/js/docs/guides/docker-images
+# You can also use any other image from Docker Hub.
+FROM apify/actor-node:18
+
+RUN apk add g++ make py3-pip
```
drobnikj, line 6: Do we really need these packages?

templates/js-langchain/.actor/actor.json
```diff
@@ -0,0 +1,15 @@
+{
+    "actorSpecification": 1,
+    "name": "project-langchain",
+    "title": "Project Langchain",
```
drobnikj, line 4:
```suggestion
    "title": "Project LangChain",
```

templates/js-langchain/.actor/actor.json
```diff
@@ -0,0 +1,15 @@
...
+    "description": "Langchain project in javascript.",
```
drobnikj, line 5:
```suggestion
    "description": "LangChain project in JavaScript.",
```

templates/js-langchain/package.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "project-langchain",
+    "version": "0.0.1",
+    "type": "module",
+    "description": "This is a boilerplate of an Apify actor.",
+    "engines": {
+        "node": ">=18.0.0"
+    },
+    "#_TODO": "Replace the langchain from Git below with an official package once our PR is merged.",
```
drobnikj, line 9: Forgoten TODO

templates/js-langchain/src/main.js
```diff
@@ -0,0 +1,86 @@
+import { Actor } from 'apify';
+import { ApifyDatasetLoader } from "langchain/document_loaders/web/apify_dataset";
```
drobnikj, line 2: I would use single quotes `'`
Maybe just lint fix the file.

templates/js-langchain/src/main.js
```diff
@@ -0,0 +1,86 @@
...
+
+// There are 2 environment variables you need to configure (https://docs.apify.com/cli/docs/vars#set-up-environment-variables-in-apify-console):
+// 1. Apify API token you obtain at https://console.apify.com/account/integrations. This is not nessesary when running this Actor at Apify platform.
+// 2. OpenAI API key you obtain at https://platform.openai.com/account/api-keys.
+const { OPENAI_API_KEY, APIFY_API_TOKEN } = process.env;
+
+// Local directory where the vector index will be stored.
+const VECTOR_INDEX_PATH = './vector_index';
+
+if (!OPENAI_API_KEY || !OPENAI_API_KEY.length) throw new Error('Please configure the OPENAI_API_KEY environment variable!');
```
drobnikj, line 28: What about t0 allowing overriding openAI token using secret input?

templates/js-langchain/src/main.js
```diff
@@ -0,0 +1,86 @@
...
+
+// Next, create the retrieval chain and enter a query:
+console.log('Asking model a question...');
+const model = new OpenAI({ openAIApiKey: OPENAI_API_KEY });
+const chain = RetrievalQAChain.fromLLM(model, vectorStore.asRetriever(), {
+    returnSourceDocuments: true,
+});
+const res = await chain.call({ query });
+
+console.log(`\n${res.text}\n`);
```
drobnikj, line 84: I would save the response into OUTPUT or dataset.

drobnikj (changes requested):
Looks good

templates/js-langchain/src/main.js
```diff
@@ -0,0 +1,86 @@
...
+
+const {
+    startUrls = [{ url: 'https://wikipedia.com' }],
+    maxCrawlPages = 3,
+    forceRecrawl = false, // Enforce a re-crawl of website content and re-creation of the vector index.
+    query = 'What is Wikipedia?',
+} = await Actor.getInput() || {};
+
+// There are 2 environment variables you need to configure (https://docs.apify.com/cli/docs/vars#set-up-environment-variables-in-apify-console):
+// 1. Apify API token you obtain at https://console.apify.com/account/integrations. This is not nessesary when running this Actor at Apify platform.
```
fnesveda, line 21: It's called `APIFY_TOKEN`, not `APIFY_API_TOKEN` on the platform.
https://github.com/apify/apify-shared-js/blob/master/packages/consts/src/consts.ts#L264

fnesveda: Honestly we should just provide both `APIFY_TOKEN` and `APIFY_API_TOKEN` on the platform, I make this mistake a lot 😄

templates/js-langchain/src/main.js
```diff
@@ -0,0 +1,86 @@
...
+// 2. OpenAI API key you obtain at https://platform.openai.com/account/api-keys.
+const { OPENAI_API_KEY, APIFY_API_TOKEN } = process.env;
+
+// Local directory where the vector index will be stored.
+const VECTOR_INDEX_PATH = './vector_index';
+
+if (!OPENAI_API_KEY || !OPENAI_API_KEY.length) throw new Error('Please configure the OPENAI_API_KEY environment variable!');
+if (!APIFY_API_TOKEN || !APIFY_API_TOKEN.length) throw new Error('Please configure the APIFY_API_TOKEN environment variable!');
```
fnesveda, line 29: I would rather direct users to `apify login`, so that they don't have to set the env var manually. If you're logged in, the `APIFY_TOKEN` env var is set for you when you do `apify run`.

templates/js-langchain/src/vector_index_cache.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Returns an ID of a key-value store with a name VECTOR_INDEX_CACHE_STORE_NAME.
+ *
+ * @returns {String}
+ */
+async function getVectorIndexCacheStore() {
+    const { id: vectorIndexCacheStoreId } = await client.keyValueStores().getOrCreate(VECTOR_INDEX_CACHE_STORE_NAME);
+
+    return client.keyValueStore(vectorIndexCacheStoreId);
```
fnesveda, line 30: Why not just `Actor.openKeyValueStore(VECTOR_INDEX_CACHE_STORE_NAME)`? Or if you want to have it on the platform, use `{ forceCloud: true )`, but I think it's better to save it locally anyway.

templates/js-langchain/.gitignore
```diff
@@ -0,0 +1,12 @@
...
+.DS_Store
+.idea
+node_modules
+storage
+
+# This location is used to locally cache
+vector_index
+
+# TODO: Remove this once our langchain PR is merged
+# langchainjs
```
fnesveda, line 12: Forgotten TODO

fnesveda (changes requested):
I think the whole `APIFY_API_TOKEN` env var logic is too complicated and would turn people off, and it's not necessary 99% of the time.

And I would cache the index to a local key-value store rather than to the platform every time.

fnesveda:
Ping @mtrunkat, any progress on this?

drobnikj:
@mtrunkat, maybe @jirimoravcik or @barjin can finish this.

mtrunkat:
Thanks, guys for your comments. A bit shame it took so long but I finally gave it another look.

templates/js-langchain/.actor/Dockerfile
mtrunkat: Yeah, it is

mtrunkat:
@jbartadev I know you worked on the templates. Is there anything different now that I should reflect in this template?

jbartadev:
> @jbartadev I know you worked on the templates. Is there anything different now that I should reflect in this template?

@mtrunkat Yes, there is.

1. You need to add the template to the `templates/manifest.json` so it's available (and displayed) in the template library and in CLI. The position (order of how templates are displayed) is defined by this file. You need to define the category (language) and included technologies there too so it's displayed in the template detail in the console and on the web.

2. We have a specific `README` structure to follow - there are keywords and patterns we use. Please see for example https://github.com/apify/actor-templates/tree/master/templates/js-start or an [example structure in Notion](https://www.notion.so/apify/Template-content-43b23aa9cae64ff8a2b286284297c1a5).

Please let us know if you need any help with it.

templates/js-langchain/src/vector_index_cache.js
```diff
@@ -0,0 +1,51 @@
...
+ * @returns {Boolean} noting if the vector index was found in the cache
+ */
+export async function retrieveVectorIndex(config) {
+    const vectorIndexCacheStore = await Actor.openKeyValueStore(VECTOR_INDEX_CACHE_STORE_NAME);
+
+    const vectorIndexRecord = await vectorIndexCacheStore.getValue(getIndexCacheKey(config));
+    if (!vectorIndexRecord) return false;
+
+    const tempFilePath = await tempWrite(vectorIndexRecord);
+    await tar.x({ file: tempFilePath, strip: 1, C: '.' });
```
fnesveda, line 48: You don't need the temporary file here, you could just do something like `await stream.finished(Readable.from(vectorIndexRecord).pipe(tar.x(...)))`

fnesveda: And the writing you could do with something like https://www.npmjs.com/package/concat-stream instead of the temporary files

fnesveda (changes requested):
The caching logic is quite complicated, with the temporary files and everything, I think it could be simplified quite a bit.

templates/js-langchain/src/vector_index_cache.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @param {Object} config
+ * @param {String} indexPath
+ */
+export async function cacheVectorIndex(config, indexPath) {
+    const vectorIndexCacheStore = await Actor.openKeyValueStore(VECTOR_INDEX_CACHE_STORE_NAME);
+    const tempFilePath = await tempWrite(tar.c({}, [indexPath]));
+    const gzipedVectorIndex = await readFile(tempFilePath);
+
+    await vectorIndexCacheStore.setValue(getIndexCacheKey(config), gzipedVectorIndex, { contentType: 'application/tar' });
```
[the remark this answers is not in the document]
mtrunkat: Yeah, true I removed the gziping as we gzip in kv-store anyway so it was pointless. I will update the naming.

templates/js-langchain/src/vector_index_cache.js
mtrunkat: This is the thing. I originally went for this approach, but I thought the Apify SDK (`getValue`/`setValue`) does not have streaming support. But it does, you just need to tell the content type ahead 🤦 so everything is like 10x simpler now.

You don't even need any extra package, just core ones:

```
import { finished } from 'node:stream/promises';

await finished(Readable.from(vectorIndexRecord).pipe(tar.x({ strip: 1, C: '.' })));
```

```
const packedVectorIndexStream = tar.c({}, [indexPath]);

await vectorIndexCacheStore.setValue(getIndexCacheKey(config), packedVectorIndexStream, { contentType: 'application/tar' });
```

drobnikj (approved):
LGTM

templates/manifest.json
```diff
@@ -360,6 +360,27 @@
...
+            "id": "js-langchain",
+            "name": "project_langchain_js",
+            "label": "Langchain",
+            "category": "javascript",
+            "technologies": [
+                "nodejs",
+                "langchain"
+            ],
+            "description": "Example of how to use LangChain.js with Apify to crawl the web data, vectorize them, and prompt the OpenAI model.",
+            "archiveUrl": "https://github.com/apify/actor-templates/blob/master/dist/templates/js-crawlee-puppeteer-chrome.zip?raw=true",
```
fnesveda, line 374:
```suggestion
            "archiveUrl": "https://github.com/apify/actor-templates/blob/master/dist/templates/js-langchain.zip?raw=true",
```

templates/js-langchain/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "dependencies": {
+        "apify": "^3.0.0",
+        "hnswlib-node": "^1.4.2",
+        "langchain": "^0.0.82",
+        "tar": "^6.1.14"
+    },
+    "scripts": {
+        "start": "node src/main.js",
+        "lint": "./node_modules/.bin/eslint ./src --ext .js,.jsx",
+        "lint:fix": "./node_modules/.bin/eslint ./src --ext .js,.jsx --fix",
```
fnesveda, line 18: There's no `eslint` in `devDependencies` here

templates/js-langchain/src/main.js
```diff
@@ -0,0 +1,98 @@
...
+    forceRecrawl = false, // Enforce a re-crawl of website content and re-creation of the vector index.
+    query = 'What is Wikipedia?',
+    openAIApiKey = OPENAI_API_KEY, // This is a fallback to the OPENAI_API_KEY environment variable when value is not present in the input.
+} = await Actor.getInput() || {};
+
+// Local directory where the vector index will be stored.
+const VECTOR_INDEX_PATH = './vector_index';
+
+if (!openAIApiKey || !openAIApiKey.length) throw new Error('Please configure the OPENAI_API_KEY as environment variable or enter it into the input!');
+if (!APIFY_TOKEN || !APIFY_TOKEN.length) throw new Error('Please configure the APIFY_TOKEN environment variable! Call `apify login` in your terminal to authenticate.');
```
fnesveda, line 32: Empty strings are `false`-y in JavaScript already, no need to check the `length`
```suggestion
if (!openAIApiKey) throw new Error('Please configure the OPENAI_API_KEY as environment variable or enter it into the input!');
if (!APIFY_TOKEN) throw new Error('Please configure the APIFY_TOKEN environment variable! Call `apify login` in your terminal to authenticate.');
```

fnesveda (changes requested):
Looks good, last few notes

PR #146, opened 2023-05-24
feat: Reverting dynamic READMEs generation with template structure

jbartadev:
We decided to go back to static READMES as we want to make each one of them specific to the template. With the current structure, it doesn't have any value to the dev as there is no information about what the template does and how to start with it, adjust it, or go to the right resources.

We'll start updating all the READMES, and comments in the code and even change the code to be more use-case based with this epic https://github.com/apify/apify-web/issues/2662


fnesveda (approved):

mnmkng (approved):
Installation commands will now be handled by UI

PR #147, opened 2023-05-24
feat: add empty python template

HonzaTuron:
Empty Python template with default docker, simple actor.json and one dependency.


templates/python-empty/README.md
```diff
@@ -0,0 +1,15 @@
+# Empty Python Actor template
+
+The Empty Python Actor template is a starting point for creating a new Python actor on the Apify platform. It provides a basic structure for the actor and allows you to easily add your own functionality.
```
mnmkng, line 3:
```suggestion
The Empty Python Actor template is a starting point for creating a new Python Actor on the Apify platform. It provides a basic structure for the Actor and allows you to easily add your own functionality. 
```

templates/python-empty/README.md
```diff
@@ -0,0 +1,15 @@
+# Empty Python Actor template
+
+The Empty Python Actor template is a starting point for creating a new Python actor on the Apify platform. It provides a basic structure for the actor and allows you to easily add your own functionality. 
+
+## Getting Started
+
+Start a new web scraping project quickly and easily in Python with our empty project template. It provides a basic structure for the Actor with [Apify SDK](https://docs.apify.com/sdk/python/) and allows you to easily add your own functionality.
```
mnmkng, line 7: Isn't this a duplicate of the first paragraph?

mnmkng (changes requested):

templates/python-empty/README.md
HonzaTuron: So should I remove it?

templates/python-start-scrape-single-page/src/main.py
```diff
@@ -0,0 +1,35 @@
...
+                soup = BeautifulSoup(response.content, 'html.parser')
+
+                # Push the title of the page into the default dataset
+                title = soup.title.string if soup.title else None
+                await Actor.push_data({ 'url': url, 'title': title })
+            except:
+                Actor.log.exception(f'Cannot extract data from {url}.')
+            finally:
+                # Mark the request as handled so it's not processed again
+                await default_queue.mark_request_as_handled(request)
```
fnesveda, line 35: I don't understand this template at all. If it's supposed to scrape a single page, then it doesn't need to use the request queue at all, no?

Plus some of the imports and variables are unused, it breaks if it gets no input...

templates/python-start-scrape-single-page/.actor/actor.json
```diff
@@ -0,0 +1,42 @@
+{
+    "actorSpecification": 1,
+    "name": "getting-started-python-beautifulsoup",
+    "title": "Getting started with Python and BeautifulSoup",
```
fnesveda, line 4: Is this the right name and title?

templates/manifest.json
```diff
@@ -1,5 +1,39 @@
...
+            "defaultRunOptions": {
+                "build": "latest",
+                "memoryMbytes": 1024,
+                "timeoutSecs": 3600
+            },
+            "showcaseFiles": [
+                "src/main.py",
+                "src/__main__.py"
+            ]
+        },
```
fnesveda, line 36: These should probably not be the first Python templates, but rather the last ones, same as how @jbartadev put the `js-empty` template as last of the JS templates in https://github.com/apify/actor-templates/pull/148.

fnesveda (changes requested):

templates/python-start-scrape-single-page/src/main.py
```diff
@@ -0,0 +1,24 @@
...
+
+import requests
+from apify import Actor
+from bs4 import BeautifulSoup
+
+async def main():
+    async with Actor:
+        # Read the Actor input
+        actor_input = await Actor.get_input() or {}
+        url = actor_input.get('start_url')
```
jbartadev, line 11: Let's unify the naming - it should be just `url` - it is not just starting the crawling with this URL, it scrapes the data only from this one specific URL in this use case

templates/python-start-scrape-single-page/src/main.py
```diff
@@ -0,0 +1,24 @@
...
+
+        Actor.log.info(f'Scraping {url}...')
+
+        try:
+            # Fetch the URL using `requests` and parse it using `BeautifulSoup`
+            response = requests.get(url)
+            soup = BeautifulSoup(response.content, 'html.parser')
+
+            # Push the title of the page into the default dataset
+            title = soup.title.string if soup.title else None
```
jbartadev, line 21: We want to have the logic unified, let's extract all the headings instead of just the title - we want the to user see more items saved to the dataset and to be logged out.

Please check the PR for js https://github.com/apify/actor-templates/pull/150/files and
1. Update the logic to have the same output
2. update the comments to follow the same structure

templates/python-start-scrape-single-page/src/__main__.py
```diff
@@ -0,0 +1,20 @@
+import asyncio
```
jbartadev, line 1: Do we really need the logger here? The default console log should be enough for this use case. I don't think we need to use the Apify one

templates/python-empty/src/main.py
```diff
@@ -0,0 +1,6 @@
+from apify import Actor
+
+
+async def main():
+    async with Actor:
+        print('Hello world!')
```
jbartadev, line 6:
```suggestion
        print('Hello from the Actor!')
```

jbartadev (changes requested):
A couple of comments before we prepare the final format of the README

templates/python-start-scrape-single-page/src/__main__.py
fnesveda: The SDK uses the logger internally, if this is not there, its log will be missing or ugly.

Plus I expect that people will want to use this template to create more complicated actors too, it would be weird if they don't have the logger setup. 

And this should be just an internal file that the users shouldn't need to touch anyway, that's why the logger complexity is hidden here instead of the main `main.py` file.

templates/python-start-scrape-single-page/src/main.py
```diff
@@ -0,0 +1,29 @@
...
+        for heading in soup.find_all(['h1', 'h2', 'h3', 'h4', 'h5', 'h6']):
+            heading_object = { 'level': heading.name, 'text': heading.text }
+            print('Extracted heading', heading_object)
+            headings.append(heading_object)
+
+        # Save headings to Dataset - a table-like storage.
+        await Actor.push_data(headings)
+
+        # Gracefully exit the Actor process. It's recommended to quit all Actors with an exit().
+        await Actor.exit()
```
fnesveda, line 29: Missing final newline, do you have EditorConfig turned on?

templates/python-start-scrape-single-page/README.md
```diff
@@ -0,0 +1,58 @@
+# BeautifulSoup & Requests template
+
+Example of a web scraper that uses Python Requests to scrape HTML from URLs provided on input, parses it using BeautifulSoup, gets all headings of the page and saves results to the storage.
+
```
fnesveda, line 4:
```suggestion
Example of a web scraper that uses Python Requests to scrape HTML from a URL provided on input, parses it using BeautifulSoup, gets all headings of the page and saves results to the storage.

```

templates/python-start-scrape-single-page/.actor/actor.json
```diff
@@ -0,0 +1,42 @@
...
+                                "format": "text"
+                            },
+                            "title": {
+                                "label": "Title",
+                                "format": "text"
+                            }
+                        }
+                    }
+                }
+            }
```
fnesveda, line 39: This does not reflect the output structure

templates/python-start-scrape-single-page/.actor/actor.json
```diff
@@ -0,0 +1,42 @@
+{
+    "actorSpecification": 1,
+    "name": "python-start-scrape-single-page",
+    "title": "Getting started with Python and BeautifulSoup",
+    "description": "Scrapes titles of websites using BeautifulSoup.",
```
fnesveda, line 5: This does not fit what the template does

templates/manifest.json
```diff
@@ -95,6 +95,40 @@
...
+        {
+            "id": "python-start-scrape-single-page",
+            "name": "python-start-scrape-single-page",
+            "label": "Scrape single page",
+            "category": "python",
+            "description": "A simple Python example of core Actor and Apify SDK features. It reads and validates user input with schema, computes a result and saves it to storage.",
```
fnesveda, line 120: I think this could be said better, something about it scraping a single page

fnesveda (changes requested):

templates/python-start-scrape-single-page/.actor/actor.json
HonzaTuron: Removed to unify it with other scrape-single-page templates.

templates/python-start-scrape-single-page/src/main.py
fnesveda, line 29: This is not necessary in Python Actors - the `async with Actor` context manager does the `Actor.exit()` for you.
```suggestion
```

fnesveda (approved):

templates/python-start-scrape-single-page/src/main.py
```diff
@@ -0,0 +1,26 @@
+from urllib.parse import urljoin
```
fnesveda, line 1: This import is unused now

fnesveda (changes requested):
One more thing

fnesveda (approved):

templates/manifest.json
```diff
@@ -95,6 +95,40 @@
...
+                "timeoutSecs": 3600
+            },
+            "showcaseFiles": [
+                "src/main.py",
+                "src/__main__.py"
+            ]
+        },
+        {
+            "id": "python-start-scrape-single-page",
+            "name": "python-start-scrape-single-page",
```
mnmkng, line 117: Pls make sure the IDs are aligned with the other languages. AFAIK they're gonna be `language-start` everywhere, but @jbartadev knows more I guess.

jbartadev: Yop, don't worry. We'll unify it before we put it out.

PR #150, opened 2023-05-24
feat: Added `js-scrape-single-page` template

jbartadev:
Have not created the zip file and manifest yet. I will do so after PR review


templates/js-start-scrape-single-page/.actor/Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Specify the base Docker image. You can read more about
+# the available images at https://docs.apify.com/sdk/js/docs/guides/docker-images
+# You can also use any other image from Docker Hub.
+FROM apify/actor-node:16
```
mnmkng, line 4: I wonder if maybe we should take this opportunity and bump all templates to Node 18? It's LTS and 16 is pretty old already.

templates/js-start-scrape-single-page/src/main.js
```diff
@@ -0,0 +1,42 @@
+// Axios import - Promise based HTTP client for the browser and node.js (Read more at https://axios-http.com/docs/intro)
```
mnmkng, line 1:
```suggestion
// Axios - Promise based HTTP client for the browser and node.js (Read more at https://axios-http.com/docs/intro)
```

templates/js-start-scrape-single-page/src/main.js
```diff
@@ -0,0 +1,42 @@
...
+import axios from "axios";
+
+// Cheerio - The fast, flexible & elegant library for parsing and manipulating HTML and XML (Read more at https://cheerio.js.org/)
+import * as cheerio from 'cheerio';
+
+// Apify SDK - toolkit for building Apify Actors (Read more at https://docs.apify.com/sdk/js/)
+import { Actor } from "apify";
+
+// Initialize the Apify SDK
```
mnmkng, line 10:
```suggestion
// The init() call configures the Actor for its environment. It's recommended to start every Actor with an init()
```

templates/js-start-scrape-single-page/src/main.js
```diff
@@ -0,0 +1,42 @@
...
+await Actor.init();
+
+// Structure of input is defined in input_schema.json.
+const input = await Actor.getInput();
+
+// Getting the URL from input defined in ./storage/key_value_stores/default/INPUT.json
+const { url } = input;
```
mnmkng, line 17: I have two separate comments to this:

1. I don't think we should describe simple operations on variables.
2. When you create the actor in console, there won't be any storage folder, so the comment is actually misleading.

templates/js-start-scrape-single-page/src/main.js
```diff
@@ -0,0 +1,42 @@
+// Axios import - Promise based HTTP client for the browser and node.js (Read more at https://axios-http.com/docs/intro)
+import axios from "axios";
+
```
mnmkng, line 3: Maybe we can squeeze the imports together? I don't think they need to have the whitespace around them.

templates/js-start-scrape-single-page/src/main.js
```diff
@@ -0,0 +1,42 @@
...
+await Actor.init();
+
+// Structure of input is defined in input_schema.json.
+const input = await Actor.getInput();
+
+// Getting the URL from input defined in ./storage/key_value_stores/default/INPUT.json
+const { url } = input;
+
+const headings = [];
+try {
```
mnmkng, line 20: I think the try/catch is not correct here. You don't need it for the console.error as the error would get printed to the console anyway, and if you actually fail to fetch the HTML, the Actor will continue and push empty data. Which doesn't necessarily break it, but it does not seem like the correct behavior.

I'd remove it to get rid of a level of nesting and a few lines.

templates/js-start-scrape-single-page/src/main.js
```diff
@@ -0,0 +1,42 @@
...
+        });
+    });
+} catch (error) {
+    console.error(error);
+}
+
+// Pushing the data to dataset
+await Actor.pushData(headings);
+
+// Exit successfully
```
mnmkng, line 41:
```suggestion
// Gracefully exit the Actor process. It's recommended to quit all Actors with an exit()
```

templates/js-start-scrape-single-page/src/main.js
```diff
@@ -0,0 +1,42 @@
...
+        headings.push({
+            level: $(element).prop("tagName").toLowerCase(),
+            text: $(element).text(),
+        });
+    });
+} catch (error) {
+    console.error(error);
+}
+
+// Pushing the data to dataset
```
mnmkng, line 38:
```suggestion
// Save headings to Dataset - a table-like storage. https://docs.apify.com/platform/storage/dataset
```

templates/js-start-scrape-single-page/src/main.js
```diff
@@ -0,0 +1,42 @@
...
+
+// Getting the URL from input defined in ./storage/key_value_stores/default/INPUT.json
+const { url } = input;
+
+const headings = [];
+try {
+    // Fetching the HTML content of the page
+    const response = await axios.get(url);
+
+    // Load the HTML to cheerio
```
mnmkng, line 24:
```suggestion
    // Parse the downloaded HTML with Cheerio to enable data extraction.
```

mnmkng (changes requested):
Looks good. Tried to improve some of the comments and left a few nitpicks.

Got one important question though. Why create a new template instead of replacing the previous `js-start` when we know that the old one kinda sucks?

templates/js-start-scrape-single-page/.actor/Dockerfile
B4nan: I was thinking the same, v16 is going EOL in 5 months.

templates/js-scrape-single-page/main.js
```diff
@@ -10,6 +10,7 @@ await Actor.init();
 
 // Structure of input is defined in input_schema.json
 const input = await Actor.getInput();
+if (!input) throw new Error("Input is missing!");
```
mnmkng, line 13: Maybe we can use an input schema default instead of this .... redundant line? 😅

templates/manifest.json
```diff
@@ -121,22 +121,24 @@
             ]
         },
         {
-            "id": "js-start",
-            "name": "getting_started_node",
-            "label": "Start with JavaScript",
-            "description": "A simple JavaScript example of core Actor and Apify SDK features. It reads and validates user input with schema, computes a result and saves it to storage.",
+            "id": "js-scrape-single-page",
```
mnmkng, line 124: Why not keep the `js-start` ID and Start with JavaScript name? That was kinda the point, that the "primary start template" would be named `start` so that everyone understands it's the getting started example. How are the users gonna know what they should start with now?

templates/manifest.json
```diff
@@ -121,22 +121,24 @@
...
-            "archiveUrl": "https://github.com/apify/actor-templates/blob/master/dist/templates/js-start.zip?raw=true",
+            "archiveUrl": "https://github.com/apify/actor-templates/blob/master/dist/templates/js-scrape-single-page.zip?raw=true",
             "defaultRunOptions": {
                 "build": "latest",
                 "memoryMbytes": 256,
                 "timeoutSecs": 360
             },
             "showcaseFiles": [
-                "src/main.js"
+                "main.js"
```
mnmkng, line 141: I think we (me, lukas, martin, some other ppl on Slack) agreed some time ago that all code will always be in `src`

jbartadev: @mnmkng I thought it was only if there are more source files, not just the main one. Is this really what you wanted?

templates/js-scrape-single-page/main.js
jbartadev: It's because of TS complains - to keep the js template exactly the same as the ts one.

mnmkng: Yeah, thought so. What about typing the input instead of this check?

templates/manifest.json
mnmkng: @jbartadev Any news on this? I feel quite strongly about this one. We should have exactly one "getting started" template for each language. It's standard in all documentation and tutorials incl. Vercel (I know you take inspiration from it)
CleanShot 2023-05-31 at 11 08 24@2x

jbartadev: Yeah, sorry I forget to reply here. I agree with you for sure. We just stopped with the work on the template until we have figured out the readme structure and logic. I'll update it.

templates/js-scrape-single-page/main.js
jbartadev: All right, I'll remove the check.

PR #151, opened 2023-05-24
Fix tests

fnesveda:
I originally wanted to do a simple change to migrate the secrets used in the tests in CI to use organization-level secrets instead of repo-level ones. But when I was doing that change, I found out the template tests don't work at all, so I got into a very deep rabbit hole and I had to rewrite them basically from scratch.

This:
- migrates the secret used in CI to be sourced from organization-level secrets
- updates packages
- removes some old dead code
- adds linting to CI
- rewrites template tests to actually work
	- installs dependencies for both Node and Python templates
	- runs the template and actually checks if the run succeeded or not
- runs the tests in CI on Windows and Linux, with various Node and Python versions

There are still some issues:
- I had to disable linting of templates, because their lint was broken, I will fix it later when there aren't changes being made to the templates, to prevent conflicts
- Apify CLI returns exit code `0` even if the command fails, so it can't be relied on yet, I will fix it there and then update the tests here


mnmkng (approved):
Looks good. I trust that it works now 😄

drobnikj (approved):
👏

jbartadev (approved):
Nice, good job

PR #152, opened 2023-05-25
feat(ts-scrape-single-page): starter template for scraping a single page

nguyeda1:
closes apify/apify-web#2659

Have not created the zip file and manifest yet. I will do so after PR review


templates/ts-scrape-single-page/package.json
```diff
@@ -1,20 +1,25 @@
...
-        "apify": "^3.0.0"
+        "apify": "^3.1.4",
+        "axios": "^1.4.0",
+        "cheerio": "^1.0.0-rc.12"
     },
     "devDependencies": {
         "@apify/tsconfig": "^0.1.0",
         "ts-node": "^10.9.1",
-        "typescript": "^4.9.5"
+        "typescript": "^5.0.4"
```
B4nan, line 17: We might want to add `@types/node` to all the TS templates, IIRC some mapped types are not working correctly without that. But I guess that can be a separate PR narrowing it in all the templates.

templates/ts-scrape-single-page/package.json
```diff
@@ -1,20 +1,25 @@
...
     },
     "scripts": {
         "start": "npm run start:dev",
         "start:prod": "node dist/main.js",
-        "start:dev": "ts-node-esm -T src/main.ts",
+        "start:dev": "ts-node-esm -T main.ts",
```
B4nan, line 22: Better to keep the `src` folder. Currently, the tsconfig will only include files from `src`, so `main.ts` in root will fail most probably (maybe not via ts-node, but during compilation)

nguyeda1: Perhaps we could just include the `main.ts` instead of the src in `tsconfig`. We wanted to keep the structure consistent with other templates, where the main file is in the root if there are no other code files.

B4nan: I'd say it brings bad DX, you will end up with new files in the root directory after compilation (at least the `.js` and `.d.ts`, but probably even more files, there can be source maps for both of those generated files).

All the crawlee templates are using `src` folder, both JS and TS, so if you want consistency, why are you removing the folder it in the first place?

cc @mnmkng @jbartadev, I saw similar discussion in the other PR

nguyeda1: Do you mean the created files after running `tsc`? Those end up in a dist folder so it should not really affect the DX no?

mnmkng: I trust @B4nan on this. It seems the marginal simplification of the template thanks to skipping the `src` folder can actually bring important problems down the road. Looks like keeping the folder will actually make it easier for users.

nguyeda1: Fair enough thx for explanation :)

templates/ts-scrape-single-page/main.ts
```diff
@@ -0,0 +1,40 @@
...
+
+// The init() call configures the Actor for its environment. It's recommended to start every Actor with an init().
+await Actor.init();
+
+interface Input {
+    url: string;
+}
+// Structure of input is defined in input_schema.json
+const input = await Actor.getInput<Input>();
+if (!input) throw new Error("Input is missing!");
```
nguyeda1, line 16: @B4nan Is there a better way to go around this than throwing an Error? This works as well
```
const input = await Actor.getInput() as Input;
```
But I am not sure which one is better.

B4nan: I think we should have a method that throws when the input is not there in the actor sdk, it's quite common practice, but since we don't have one, I would keep it the way it is now, with the explicit throw. Better than lying to the compiler (which you would do with the type assertion).

I can add `Actor.getInputOrThrow()` there if you want, it makes total sense to me (actually I would be fine throwing from the `getInput` directly, but that's quite breaking).

nguyeda1: Adding the `getInputOrThrow` would probably require a lot of udpate in documentation as well right? I think for now I will just throw the error :) thx

B4nan: Not really, just a new method, with a jsdoc, we don't need to update existing docs because of that (at least not immediately). I will add that anyway, I wanted to do that for quite some time.

mnmkng: But we don't need the throw if we type the input, no? If we say that the input is always an object with a url: string, then we can delete the line with the exception.

I'd say that's a better tutorial pattern for a TS template than validation. Tutorials and examples usually skip validation, because if you validate one thing, then you should validate everything, add error handling and what not, so it's typical to just follow the happy path in tutorials to keep them clean and succinct.

B4nan: > But we don't need the throw if we type the input, no? If we say that the input is always an object with a url: string, then we can delete the line with the exception.

But it is not always an object, if the INPUT key is missing from the KVS, it returns `null`. And that is what we want to deal with. I guess your point of view is that the INPUT key will be part of the template, so it will be always there? Still, users can wipe the storage manually, and their code would start failing with some reference error etc, throwing is the cleaner approach.

So yes, we can assert the type, but it's still lying to the compiler, unlike conditional throwing, which is a runtime statement (that propagates to the type level).

mnmkng: Yeah, I know how it works and you also got my point correctly.

Tutorials are always "incomplete". You never account for all the possible scenarios, because that's not the point. The point is to showcase functionality in the smallest possible example that includes only what it needs to and nothing else.

Plus, even the deletion of storage directory is only a temporary problem, because after we support input schema prefills / defaults in CLI, the URL will always have a value.

But I don't have a strong opinion about it, I just think it doesn't look good and adds unnecessary noise. I never put validation in any tutorials / examples, unless it actually showcases something I'm writing about.

PR #154, opened 2023-06-07
feat: adding readme suffixes to the manifest

nguyeda1:
closes apify/apify-web#2708


templates/consoleReadmeSuffix.md
```diff
@@ -0,0 +1,39 @@
...
+    ```
+    apify pull <TITLE>
+    ```
+
+## Documentation reference
+
+To learn more about Apify and Actors, take a look at the following resources:
+
+- [Apify SDK for JavaScript documentation](https://docs.apify.com/sdk/js)
+- [Apify SDK for Python documentation](https://docs.apify.com/sdk/python)
```
nguyeda1, line 37: Added the Python documentation here in case the template is python-based. This could be dynamic somehow as well but don't think it is worth the effort

src/index.js
```diff
@@ -1,8 +1,32 @@
...
+ * This will fetches the readme suffixes for the console and local templates.
+ * These suffixes are then added to the manifest object at the root level
+ */
+async function fetchReadmeSuffixes() {
+    let consoleReadmeSuffix = '';
+    let localReadmeSuffix = '';
+
+    try {
+        const consoleResponse = await fetch(CONSOLE_README_SUFFIX_URL);
+        const localResponse = await fetch(LOCAL_README_SUFFIX_URL);
```
mnmkng, line 17: We can't use `fetch` unless we bump CLI's requirement to Node 18, which I don't think we should do.

CleanShot 2023-06-07 at 14 18 06@2x

B4nan: We could use https://github.com/nodejs/undici

src/index.js
```diff
@@ -1,30 +1,32 @@
...
+
+        if (localResponse.status === 200) {
+            manifestData.localReadmeSuffix = await localResponse.text();
+        }
+
+        return manifestData;
+    } catch (err) {
+        console.error(err);
+        throw err;
+    }
```
mnmkng, line 29: Is this needed? Wouldn't it only print the error twice and behave the same as without the try catch block?

mnmkng (changes requested):

templates/consoleReadmeSuffix.md
```diff
@@ -0,0 +1,39 @@
...
+    **Using NPM**
+
+    ```
+    npm -g install apify-cli
+    ```
+
+2. Pull the Actor by its unique `<TITLE>`. This command will copy the actor into the current directory on your local machine.
+
+    ```
+    apify pull <TITLE>
```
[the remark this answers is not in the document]
jbartadev: We'd need to parse the md and replace the `<TITLE>` with the actor name in the console. Also, it's not a title, it is a name or ID (e.g. `apify/hello-world` or `E2jjCZBezvAZnX8Rb`)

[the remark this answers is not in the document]
nguyeda1: Not sure which script you are referring to but I don't know any that would insert the actor's name into the README. 

Actors created from templates have default names - **my-actor-X** so in theory we can add it into the readme. However, the user can and will most likely rename the actor and we would have to update the README again which we would like to avoid.

So for now we would just keep the generic placeholder `<TITLE>`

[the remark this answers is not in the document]
jbartadev: I know we wanted to make it static to avoid the issues with renaming etc., so it's more for Ondra. I'm fine with keeping static `<ID>` as it's definitely a 90/10 solution here..

templates/consoleReadmeSuffix.md
```diff
@@ -0,0 +1,39 @@
...
+    brew install apify/tap/apify-cli
+    ```
+
+    **Using NPM**
+
+    ```
+    npm -g install apify-cli
+    ```
+
+2. Pull the Actor by its unique `<ID>`. This command will copy the actor into the current directory on your local machine.
```
jbartadev, line 26:
```suggestion
2. Pull the Actor by its unique `<ID>`. This command will copy the Actor into the current directory on your local machine.
```

templates/consoleReadmeSuffix.md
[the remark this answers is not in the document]
nguyeda1: Hmm I previously thought that the pull command would not work with ID for some reason. If it works with ID then we do not have to worry about the renaming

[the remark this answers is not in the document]
jbartadev: Hmm, true. I still think we can move forward with this as it is such a small detail but it'd actually make sense to dynamically add the ID in the console. Should be easy right?

package.json
```diff
@@ -45,5 +45,8 @@
     "mustache": "^4.2.0",
     "playwright": "^1.34.2",
     "puppeteer": "^20.1.0"
+  },
+  "dependencies": {
+    "undici": "^5.22.1"
```
fnesveda, line 50: Do we really need this? `undici` is an additional [1.12 MB](https://www.npmjs.com/package/undici), and we're trying to keep the CLI, where this package is imported, as light as possible.

I think that for the three files that we need to download we could just take the existing code that uses `https`, wrap it in a function and use that one, no need to add another package.

And once we drop Node 16, we can start using the built-in `fetch` available since Node 18.

templates/consoleReadmeSuffix.md
```diff
@@ -0,0 +1,39 @@
+## Getting started
```
fnesveda, line 1: Let's use `snake_case` in filenames instead of `camelCase`, as per convention:
https://www.notion.so/apify/Coding-standards-editor-setup-58912bbe2c0c4ad9961ae818179bcbd9?pvs=4#9f86cbe9e80d46e197414ca1ddba1046

templates/localReadmeSuffix.md
```diff
@@ -0,0 +1,41 @@
+## Getting started
```
fnesveda, line 1: Same here with the file naming

fnesveda (changes requested):
Looks good, just small nitpicks

package.json
B4nan: Good point, given we should drop node 16 later this year anyway.

> and we're trying to keep the CLI, where this package is imported, as light as possible.

I'd say the CLI really fat already, so this note sounds a bit funny :]

```
➜  htdocs npx howfat apify-cli
Dependencies: 434
Size: 40.1mb
Files: 12424
```

The dependency graph is really crazy:

https://npm.anvaka.com/#/view/2d/apify-cli

B4nan: Just the oclif core takes almost 10mb, and it has dependency on ts-node -> the whole TS.

```
➜  htdocs npx howfat @oclif/core
Dependencies: 111
Size: 9.69mb
Files: 1609
```

fnesveda: I'm trying to get them to drop it, but they're ignoring me 😄 https://github.com/oclif/core/pull/643

templates/consoleReadmeSuffix.md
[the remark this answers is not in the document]
mnmkng: I'm glad we replaced `<TITLE>` with <ID> as Actors don't have a title. But still, the templates are usually the first point of contact for users, so they have no idea where to find the ID and it creates friction. The readme does not even say where to find it.

Please either make it clear where and how the dev should get the ID, or replace it dynamically. In the current state, it's not helpful for new users.

templates/consoleReadmeSuffix.md
mnmkng: Lol, good catch. I prefer dash case, but we can universaly agree that camel case is wrong for file names 😅

templates/consoleReadmeSuffix.md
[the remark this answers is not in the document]
nguyeda1: The dynamic way is more tempting there is just one tiny "problem". We need the actorId for it which means:
1. Create the actor
2. Use the actorId to generate console readme
3. Update the actor

And we are not sure if the third step is okay because it might be updating the actor stats as well.

On the other hand, if we went with the static way, right now I don't see any easy way to navigate the user to find the actor ID aside from the URL.

I will discuss this today with @mtrunkat but I would be grateful for any suggestions @mnmkng :)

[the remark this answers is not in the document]
nguyeda1: So after the discussion with @mtrunkat we will go with the static way. In short, we do not want to put the ID of the actor into the readme in case someone else uses the same source code to create their new actors.

src/index.js
```diff
@@ -19,12 +21,27 @@ exports.fetchManifest = async () => {
...
+exports.fetchManifest = async () => {
+    const manifest = await fetchResource(MANIFEST_URL);
+
+    try {
+        const consoleReadmeSuffix = await fetchResource(CONSOLE_README_SUFFIX_URL);
+        const localReadmeSuffix = await fetchResource(LOCAL_README_SUFFIX_URL);
+        manifest.consoleReadmeSuffix = consoleReadmeSuffix;
+        manifest.localReadmeSuffix = localReadmeSuffix;
+    } catch (e) {
+        // Could not fetch the suffixes
```
nguyeda1, line 41: Not sure how bad of an idea this empty catch block is. Basically, the idea is that the manifest still returns from this function correctly even if the suffixes fail to load.

jbartadev, line 41: I'd probably just log it here in case we'd need to debug it.

```suggestion
        // Could not fetch the suffixes
        console.error('Could not fetch the suffixes.', e);
```

HonzaTuron (approved):

src/index.js
```diff
@@ -19,12 +21,27 @@ exports.fetchManifest = async () => {
...
         }).on('error', (err) => reject(err));
     });
 };
 
+exports.fetchManifest = async () => {
+    const manifest = await fetchResource(MANIFEST_URL);
+
+    try {
+        const consoleReadmeSuffix = await fetchResource(CONSOLE_README_SUFFIX_URL);
+        const localReadmeSuffix = await fetchResource(LOCAL_README_SUFFIX_URL);
```
fnesveda, line 37: This won't work, the `fetchResource` currently tries to parse the result as JSON, and throws if it fails.

The `fetchResource` should just return the response body as string, and then we'd either parse it as JSON in case of the manifest, or just keep it in case of these Markdown files.

fnesveda (changes requested):

src/index.js
nguyeda1: Oh thx complete forgot about that

fnesveda (approved):

HonzaTuron:
FYI you can use node-fetch instead https as it's already in package-lock.

fnesveda:
> FYI you can use node-fetch instead https as it's already in package-lock.

Yeah but only in dev dependencies, this package has no production dependencies

templates/console_readme_suffix.md
```diff
@@ -0,0 +1,43 @@
...
+    ```
+
+2. Pull the Actor by its unique `<ACTID>`, which is:
+    - name of the Actor to pull (e.g. "apify/hello-world")
+    - or ID of the Actor to pull (e.g. "E2jjCZBezvAZnX8Rb")
+
+    This command will copy the Actor into the current directory on your local machine.
+
+    ```
+    apify pull <ACTID>
```
mnmkng, line 33: It's not ACTID, but Actor ID. The fact that that it's named `actId` on the API is simply because 5 years ago Actors were called Acts, but we haven't been using this name since. ACTID is not a short for (unique name OR id).

Please guide the users where to get the ID / **unique** name (yes, not a name, we have title, unique name and ID) or add a screenshot please. This is still not good enough for the users.

CleanShot 2023-06-14 at 11 28 50@2x

CleanShot 2023-06-14 at 11 29 56@2x

mnmkng (changes requested):

templates/console_readme_suffix.md
nguyeda1: Ok ok thx I will try to add the instructions. Unforunately the screenshots are be displayed in the Web IDE readme

mnmkng (approved):
Thanks, much more helpful now 👍

jbartadev (approved):

B4nan (approved):

PR #171, opened 2023-07-26
fix(ts-start): Fixed typo .gitingore -> .gitignore

JJetmar:
Fixed based on https://apifier.slack.com/archives/C0L33UM7Z/p1690321589739739


jbartadev (approved):

HonzaTuron (approved):

fnesveda:
I think that's my fault, I thought that if you specify `chromedriver-version` for the `setup-chromedriver` action, it will install Chromedriver and Chrome in compatible versions, but it installs the specified Chromedriver version, but latest Chrome, which are not compatible.

I'll fix it in another PR, this one is good to go even with the failing tests.

PR #172, opened 2023-07-26
feat: Crawlee + Cheerio templates code refactoring + Readme update

HonzaTuron:
Closes apify/apify-web#2669.

Removing routes.js, updating readme, code refactoring, correct typing.


templates/manifest.json
```diff
@@ -301,7 +301,7 @@
                 "cheerio"
             ],
             "description": "Skeleton project that helps you quickly bootstrap `CheerioCrawler` in JavaScript. It's best for developers who already know Apify SDK and Crawlee.",
-            "archiveUrl": "https://github.com/apify/actor-templates/blob/master/dist/templates/ts-crawlee-cheerio.zip?raw=true",
+            "archiveUrl": "https://github.com/apify/actor-templates/blob/master/dist/templates/js-bootstrap-cheerio-crawler.zip?raw=true",
```
HonzaTuron, line 304: This seems to be broken for a while 😄

templates/js-crawlee-cheerio/.actor/input_schema.json
```diff
@@ -10,9 +10,15 @@
             "editor": "requestListSources",
             "prefill": [
                 {
-                    "url": "https://apify.com"
+                    "url": "https://crawlee.com"
```
jbartadev, line 13: I think this is not our domain 😄 
```suggestion
                    "url": "https://crawlee.dev"
```

templates/js-crawlee-cheerio/src/main.js
```diff
@@ -7,19 +7,27 @@
...
 // For more information, see https://crawlee.dev
-import { CheerioCrawler } from 'crawlee';
-import { router } from './routes.js';
+import { CheerioCrawler, Dataset } from 'crawlee';
 
 // Initialize the Apify SDK
 await Actor.init();
 
-const startUrls = ['https://apify.com'];
+const { startUrls =  ["https://crawlee.com"], maxRequestsPerCrawl = 100 } = await Actor.getInput();
```
jbartadev, line 15: We don't need the defaults here, the input schema has it and locally we'll also have the `INPUT.json` prefilled with defaults. Let's just take the `startUrls` and `maxRequestsPerCrawl` from the input and use it.

templates/js-crawlee-cheerio/src/main.js
```diff
@@ -7,19 +7,27 @@
...
 
 const proxyConfiguration = await Actor.createProxyConfiguration();
 
 const crawler = new CheerioCrawler({
     proxyConfiguration,
-    requestHandler: router,
+    maxRequestsPerCrawl,
+    requestHandler: async ({ enqueueLinks, request, $, log }) => {
+        log.info('enqueueing new URLs');
+        await enqueueLinks( { globs: ['https://crawlee.com/*'], label: 'detail' });
```
jbartadev, line 24: The `globs` can't be defined like this because it wouldn't work if you change the input to a different domain. Also the `label: 'detail'` doesn't make sense anymore as the crawler doesn't use routes

templates/ts-crawlee-cheerio/.actor/input_schema.json
```diff
@@ -10,9 +10,15 @@
             "editor": "requestListSources",
             "prefill": [
                 {
-                    "url": "https://apify.com"
+                    "url": "https://crawlee.com"
```
jbartadev, line 13:
```suggestion
                    "url": "https://crawlee.dev"
```

templates/js-crawlee-cheerio/README.md
```diff
@@ -1,12 +1,19 @@
-# CheerioCrawler Actor template
+# JavaScript Cheerio Actor template
```
jbartadev, line 1: I'd probably keep the `CheerioCrawler Actor template` as this template is not purely JavaScript + Cheerio library. It is using the CheerioCralwer which is defined by the Crawlee library (see [this](https://crawlee.dev/api/cheerio-crawler/class/CheerioCrawler) for more info)

I'd maybe go for something like `Crawlee CheerioCrawler Actor template` to make it even more clear

templates/js-crawlee-cheerio/README.md
```diff
@@ -1,12 +1,19 @@
...
 
-This template is a production ready boilerplate for developing with `CheerioCrawler`. Use this to bootstrap your projects using the most up-to-date code.
+A template example built with [Crawlee](https://crawlee.dev) to scrape data from a website using [Cheerio](https://cheerio.js.org/).
```
jbartadev, line 3: I think this should mention the `CheerioCrawler` too https://crawlee.dev/api/cheerio-crawler/class/CheerioCrawler

templates/manifest.json
jbartadev: Haha, nice, good catch!

jbartadev (changes requested):

templates/js-crawlee-cheerio/README.md
HonzaTuron: I'm gonna let there language TypeScript/JavaScript to avoid confusion between templates.

nguyeda1 (commented):
Looks good but the tests are failing :/

jbartadev:
> Looks good but the tests are failing :/

@nguyeda1 @HonzaTuron 
I think it's not the fault of this PR, Franta mentioned it here
https://github.com/apify/actor-templates/pull/171#issuecomment-1651853087

nguyeda1 (approved):

templates/ts-crawlee-cheerio/README.md
```diff
@@ -1,12 +1,19 @@
...
-- [Cheerio Tutorial](https://crawlee.dev/docs/guides/cheerio-crawler-guide)
-- [Documentation](https://crawlee.dev/api/cheerio-crawler/class/CheerioCrawler)
-- [Examples](https://crawlee.dev/docs/examples/cheerio-crawler)
+## How it works
+
+This code is a TypeScript script that uses Cheerio to scrape data from a website. It then stores the website titles in a dataset.
+
+- The script loads the HTML of the provided URLs from `startUrls` field in input schema.
+- Uses Cheerio `requestHandler` function to scrape the website titles.
+- Number of crawls is limited by `maxPagesPerCrawl` field from input schema.
```
jbartadev, line 18: I'd move this under the `startUrls` bullet point.

jbartadev (changes requested):
Nice, I have a couple of comments on the template description, that should be applied to both templates. 

Also, adding @B4nan as a reviewer to make sure this is the template behavior we agreed on.

templates/js-crawlee-cheerio/src/main.js
```diff
@@ -7,19 +7,27 @@
...
-const startUrls = ['https://apify.com'];
+const { startUrls, maxRequestsPerCrawl } = await Actor.getInput() || {};
 
 const proxyConfiguration = await Actor.createProxyConfiguration();
 
 const crawler = new CheerioCrawler({
     proxyConfiguration,
-    requestHandler: router,
+    maxRequestsPerCrawl,
+    requestHandler: async ({ enqueueLinks, request, $, log }) => {
```
B4nan, line 22: I'd personally rather see the ES6 shorthand syntax here, we use that in the crawlee docs.

```suggestion
    async requestHandler({ enqueueLinks, request, $, log }) {
```

less special characters => easier to read and write

templates/ts-crawlee-cheerio/src/main.ts
```diff
@@ -7,19 +7,27 @@
...
 // For more information, see https://crawlee.dev
-import { CheerioCrawler } from 'crawlee';
-import { router } from './routes.js';
+import { CheerioCrawler, Dataset, Dictionary } from 'crawlee';
 
 // Initialize the Apify SDK
 await Actor.init();
 
-const startUrls = ['https://apify.com'];
+const { startUrls, maxRequestsPerCrawl } = (await Actor.getInput() as Dictionary) || {};
```
B4nan, line 15: you should have an interface here, and pass it to the generic type arg instead of using type assertions

we should use best practices in the templates, this is rather workaround

templates/js-crawlee-cheerio/README.md
```diff
@@ -13,5 +14,5 @@ A template example built with [Crawlee](https://crawlee.dev) to scrape data from
 
 This code is a JavaScript script that uses Cheerio to scrape data from a website. It then stores the website titles in a dataset.
 
-- The crawler starts with URLs provided from the input `startUrls` field defined by the input schema. Number of crawls is limited by `maxPagesPerCrawl` field from input schema.
+- The crawler starts with URLs provided from the input `startUrls` field defined by the input schema. Number of crawler pages is limited by `maxPagesPerCrawl` field from input schema.
```
B4nan, line 17: not "crawler pages" but "crawled pages" - the pages that were crawled

templates/ts-crawlee-cheerio/src/main.ts
```diff
@@ -7,19 +7,32 @@
...
+interface Input {
+    startUrls: string[];
+    maxRequestsPerCrawl: number;
+}
 
 // Initialize the Apify SDK
 await Actor.init();
 
-const startUrls = ['https://apify.com'];
+const { startUrls, maxRequestsPerCrawl } = await Actor.getInputOrThrow<Input>();
```
B4nan, line 20: I just realized we don't have the INPUT key in the git repo, so running this locally will probably throw...

I guess let's do it the same way as in the JS template? You will need to type the `{}` too, otherwise it should look the same.

```suggestion
const { startUrls, maxRequestsPerCrawl } = await Actor.getInputOrThrow<Input>();

const {
    startUrls = ['https://crawlee.dev'],
    maxRequestsPerCrawl = 100,
} = await Actor.getInput<Input>() ?? {} as Input;
```

B4nan (approved):

PR #173, opened 2023-07-26
fix: Specify the same Chrome and Chromedriver version in tests

fnesveda:
Apparently, if you pass `chromedriver-version` to `setup-chromedriver`, it will install the given Chromedriver version, but it will also install latest Chrome which might not be compatible to the given Chromedriver version. 

There is now a better way to install the same Chrome and Chromedriver, using [Chrome for Testing](https://developer.chrome.com/blog/chrome-for-testing/) and `npx @puppeteer/browsers install ...`.

### DIFF
--- a/.github/workflows/lint_and_test.yaml
+++ b/.github/workflows/lint_and_test.yaml
@@ -124,8 +124,8 @@ jobs:
 
     - name: Setup Chromedriver
       run: |
-        npx @puppeteer/browsers install chrome@stable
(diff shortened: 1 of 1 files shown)


fnesveda:
Damn, this doesn't work, now I know why I had to specify the version there in the first place 😄

B4nan (approved):

vdusek (approved):
LGTM

PR #184, opened 2023-08-31
Python Scrapy Actor uses Request Queue

### LINKED ISSUE #183: Update Python Scrapy Actor template to use the Request Queue as storage
### Current behavior

- Python Scrapy Actor uses Scrapy's default Scheduler with its default storage.

### Expected behavior

- Python Scrapy Actor uses our custom (asynchronous) Scheduler with our Request Queue as a storage.

### Possible solutions

- The necessary investigation was already made:
    - https://github.com/apify/apify-core/issues/11884 
    - https://github.com/apify/apify-core/issues/12073.

### Additional context
    
- Apify SDK 1.1.4 should be used with this fix https://github.com/apify/apify-sdk-python/pull/113.

vdusek:
## Description

- Update the Python Scrapy Actor template to use our Request Queue (and not Scrapy default storage).
- This was quite a challenge, mainly because of the usage of the [Twisted](https://twisted.org/) library (alternative to Asyncio) in the Scrapy project.
- PR mainly provides our custom [Scheduler](https://docs.scrapy.org/en/latest/topics/scheduler.html) and a customized version of [RetryMiddleware](https://docs.scrapy.org/en/latest/_modules/scrapy/downloadermiddlewares/retry.html). Both of these components interact with Request Queue.
- All interactions with the Request Queue have to be executed outside of the main event loop. This is solved by creating a  custom nested event loop and all the RQ interactions are executed in it.

### Exceptions in `run_until_complete`

If we use `run_until_complete` in a custom-nested event loop like this:

```python
try:
    event_loop.run_until_complete(foo_coroutine())
except BaseException:
    traceback.print_exc()
```

We must wrap it into a try block and call `traceback.print_exc()` in the except part if we want the exception to be propagated. Otherwise, the request processing will be terminated without any notice in the log or something and Scrapy will just continue with the next one. Which could be pretty confusing...

### "Robots" requests

"Robots" requests (`*/robots.txt`) are bypassed directly from the Engine (through middlewares) to the Spider. They don't go through a Scheduler. It would be pretty hard to try to force them to go through the Request Queue. So in our Retry Middleware, we identify these requests and do not make any interaction with the Request Queue in such cases.
## Ticket

- Closes #183 

## Blocked by

- Merge https://github.com/apify/apify-sdk-python/pull/113 and release a new version of SDK first.


templates/python-scrapy/src/middlewares.py
```diff
@@ -0,0 +1,71 @@
...
+        except BaseException:
+            traceback.print_exc()
+
+    def __del__(self):
+        nested_event_loop.stop()
+        nested_event_loop.close()
+
+    def process_response(self, request: Request, response: Response, spider: Spider) -> Request | Response:
+        """
+        Process the response and decide whether the request should be retried.
```
B4nan, line 31: I guess this `process_response` method name comes from some predefined interface we need to adhere to? If not, I'd expect it to have a better name and a different return type (if its purpose is to tell whether a request should be retried or not).

B4nan (approved):
just one suggestion and few questions

(haven't tried it locally)

templates/python-scrapy/src/middlewares.py
vdusek: Unfortunately, this is a mandatory structure for Scrapy's [Downloader Middlewares](https://docs.scrapy.org/en/latest/topics/downloader-middleware.html).

vdusek:
> just one suggestion and few questions
> (haven't tried it locally)

Just in case, you would like to try it locally, it should be pretty easy (with default Actor input):

```
cd templates/python-scrapy/
```
```
virtualenv --python $(which python3.11) .venv
```
```
source .venv/bin/activate
```
```
pip install -r requirements.txt
```
```
apify run --purge
```

templates/python-scrapy/src/main.py
```diff
@@ -3,23 +3,32 @@
...
 
         settings = get_project_settings()
-        settings['ITEM_PIPELINES'] = { ActorDatasetPushPipeline: 1 }
+        settings['ITEM_PIPELINES'] = {'src.pipelines.ActorDatasetPushPipeline': 1}
+        settings['SCHEDULER'] = 'src.scheduler.ApifyScheduler'
+        settings['DOWNLOADER_MIDDLEWARES'] = {
+            'scrapy.downloadermiddlewares.retry.RetryMiddleware': None,
+            'src.middlewares.ApifyRetryMiddleware': 543,
+        }
         settings['DEPTH_LIMIT'] = max_depth
```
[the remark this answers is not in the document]
vdusek: Good idea, done

fnesveda (changes requested):
Looks pretty good! I have some issues with it on the platform though, it finishes while there are still some pending requests in the queue, check run ID `vSS8GqeHThObthovh`

templates/python-scrapy/requirements.txt
```diff
@@ -1,7 +1,7 @@
 # Add your dependencies here.
 # See https://pip.pypa.io/en/latest/reference/requirements-file-format/
 # for how to format them
-apify ~= 1.1.3
+apify ~= 1.1.4
 nest-asyncio ~= 1.5.7
 scrapy ~= 2.10.0
```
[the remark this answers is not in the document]
vdusek: Agreed, thanks.

templates/python-scrapy/src/main.py
```diff
@@ -1,25 +1,45 @@
...
 from scrapy.utils.project import get_project_settings
+from scrapy.settings import Settings
 
 from apify import Actor
 
-from .pipelines import ActorDatasetPushPipeline
 from .spiders.title_spider import TitleSpider
 
 
+def get_settings(max_depth: int) -> Settings:
```
fnesveda, line 10: I would maybe call it `get_scrapy_settings()` to make it clear

fnesveda (approved):
Looks good, one last naming nitpick 🙂 And the tests are already fixed in `master`, so merging that should be enough for them to pass here

PR #203, opened 2023-09-14
feat: Added bun-start template

valekjo:
This PR adds the most basic bun template.

The build is super fast.


templates/manifest.json
```diff
@@ -508,6 +508,31 @@
...
+            "technologies": [
+                "bun",
+                "cheerio",
+                "axios"
+            ],
+            "description": "Scrape single page with provided URL with Axios and extract data from page's HTML with Cheerio.",
+            "archiveUrl": "https://github.com/apify/actor-templates/blob/master/dist/templates/ts-start-bun.zip?raw=true",
+            "defaultRunOptions": {
+                "build": "latest",
+                "memoryMbytes": 256,
```
mtrunkat, line 526: Suggestion: I'd consider giving it a GB for a better user experience, it will speedup the container start likely

valekjo: It's really fast as it is.

templates/ts-start-bun/.actor/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# Specify the base Docker image
+FROM oven/bun:1.0
```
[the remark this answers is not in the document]
valekjo: The image is small (100MB) and the build overall is much faster, even like this.

mtrunkat (changes requested):

PR #204, opened 2023-09-18
chore: fix ci for new playwright

vladfrangu:


B4nan:
sidenote: where is that `0.5` estimate coming from? was it there or did you add it? i dont think we need that, `1` means its a simple task that took up to one hour, this does not need to be precise, we dont need to care about "this was even simpler than 1 hour". usually even the super simple things end up as ~1 hour due to the reviews and whatnot

vladfrangu:
Yeah that was added by me, since I've never really dealt with estimates before 😅😅, good to know

B4nan:
so the problem was version mismatch between what is used in the tests and what is used in the template itself?

vladfrangu:
Yep, that was all. Base repo installed <1.38.0 (not sure exactly which version), which also installed an older chrome, whereas the templates installed `*` -> `whatever @latest is` -> expecting newer chrome

B4nan:
In that case I'd say we should have `*` in the root dependencies too.

vladfrangu:
Not fully sure if that'll matter since the root has a lockfile.. we can *try* it and see what happens?

B4nan:
Hmm, maybe we should add renovate to this repo too, at least for the lockfile updates.

For now, let's just merge so the tests are passing again.

B4nan:
looks like it did not help at all, how come the PR checks were passing?

vladfrangu:
Well it failed for a completely different reason, seems to be related to langchain

PR #206, opened 2023-09-18
Update structure of the Scrapy Actor

### LINKED ISSUE #205: Update structure of the Scrapy Actor
Move our internal components into an `apify/` subfolder. This should make it clear to users which files they should interfere with and which they should not.

vdusek:
closes #205


templates/python-scrapy/src/apify/main.py
```diff
@@ -4,21 +4,21 @@
...
+from ..spiders.title_spider import TitleSpider
 
 
-def get_scrapy_settings(max_depth: int) -> Settings:
+def _get_scrapy_settings(max_depth: int) -> Settings:
     """
     Get Scrapy project settings.
     """
     settings = get_project_settings()
     settings['ITEM_PIPELINES'] = {
```
fnesveda, line 15: Now I'm thinking, shouldn't this actually add the pipeline to the existing `ITEM_PIPELINES` dict, rather than overwriting it?

fnesveda: Same for the `DOWNLOADER_MIDDLEWARES`, and maybe the `DEPTH_LIMIT` logic could also be improved

templates/python-scrapy/src/pipelines.py
```diff
@@ -1,18 +1,14 @@
...
 
-from itemadapter import ItemAdapter
+import scrapy
 
-from scrapy import Item, Spider
+from .items import TitleItem
 
-from apify import Actor
 
+class TitleItemPipeline:
```
fnesveda, line 10: This pipeline won't get applied unless you add it to settings

fnesveda (changes requested):
Looks good, I think there are some small settings adjustments needed though

templates/python-scrapy/src/apify/main.py
vdusek: Oh, good point... In the `settings.py`, we should create a new dictionary and Scrapy will just merge them. In this case, we should probably just update the existing one.

I checked the log with the current setup and it works for the middlewares (the default ones were still there), however, does not work for Pipelines (pipelines from settings.py were overridden by this). 

Anyway, now it seems good:

```
INFO  Overridden settings:
      {'BOT_NAME': 'titlebot',
       'DEPTH_LIMIT': 1,
       'NEWSPIDER_MODULE': 'src.spiders',
       'REQUEST_FINGERPRINTER_IMPLEMENTATION': '2.7',
       'ROBOTSTXT_OBEY': True,
       'SCHEDULER': 'src.apify.scheduler.ApifyScheduler',
       'SPIDER_MODULES': ['src.spiders'],
       'TWISTED_REACTOR': 'twisted.internet.asyncioreactor.AsyncioSelectorReactor'}
...
INFO  Enabled extensions:
      ['scrapy.extensions.corestats.CoreStats',
       'scrapy.extensions.telnet.TelnetConsole',
       'scrapy.extensions.memusage.MemoryUsage',
       'scrapy.extensions.logstats.LogStats'] ({"crawler": "<scrapy.crawler.Crawler object at 0x7fd92381bd90>"})
INFO  Enabled downloader middlewares:
      ['scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware',
       'scrapy.downloadermiddlewares.httpauth.HttpAuthMiddleware',
       'scrapy.downloadermiddlewares.downloadtimeout.DownloadTimeoutMiddleware',
       'scrapy.downloadermiddlewares.defaultheaders.DefaultHeadersMiddleware',
       'scrapy.downloadermiddlewares.useragent.UserAgentMiddleware',
       'scrapy.downloadermiddlewares.redirect.MetaRefreshMiddleware',
       'scrapy.downloadermiddlewares.httpcompression.HttpCompressionMiddleware',
       'scrapy.downloadermiddlewares.redirect.RedirectMiddleware',
       'scrapy.downloadermiddlewares.cookies.CookiesMiddleware',
       'scrapy.downloadermiddlewares.httpproxy.HttpProxyMiddleware',
       'scrapy.downloadermiddlewares.stats.DownloaderStats',
       'src.apify.middlewares.ApifyRetryMiddleware'] ({"crawler": "<scrapy.crawler.Crawler object at 0x7fd92381bd90>"})
INFO  Enabled spider middlewares:
      ['scrapy.spidermiddlewares.httperror.HttpErrorMiddleware',
       'scrapy.spidermiddlewares.offsite.OffsiteMiddleware',
       'scrapy.spidermiddlewares.referer.RefererMiddleware',
       'scrapy.spidermiddlewares.urllength.UrlLengthMiddleware',
       'scrapy.spidermiddlewares.depth.DepthMiddleware'] ({"crawler": "<scrapy.crawler.Crawler object at 0x7fd92381bd90>"})
INFO  Enabled item pipelines:
      ['src.apify.pipelines.ActorDatasetPushPipeline',
       'src.pipelines.TitleItemPipeline'] ({"crawler": "<scrapy.crawler.Crawler object at 0x7fd92381bd90>"})
```

templates/python-scrapy/src/pipelines.py
vdusek: I thought it could be just a placeholder. To let the users know where they can define their pipelines. However, we can just return the original Item and use it. Why not.

vdusek:
I also updated to Scrapy 2.11 ([release notes](https://docs.scrapy.org/en/latest/news.html#scrapy-2-11-0-2023-09-18)). It shouldn't be a problem.

fnesveda (approved):
Looks good, just one last thing - please also update the showcase files in https://github.com/apify/actor-templates/blob/scrapy-update-structure/templates/manifest.json#L135-L140 to refer to the new file locations, and then it's good to go

PR #212, opened 2023-10-02
chore: add postinstall command for playwright templates

vladfrangu:
Related: https://github.com/apify/crawlee/pull/2104


B4nan:
The tests are apparently broken, do we also need to ship new crawlee version for that? I guess we might want to enforce that new crawlee version in the templates?

vladfrangu:
Yep, templates currently fail cause of old crawlee without the CLI command

PR #213, opened 2023-10-04
Format Python Actor templates

vdusek:
U used auto-formatters (`autopep8`, `isort`) that we use in our other Python libraries (`SDK`, `Client`) to format the Python Actor templates.

I also used linter (`flake8` with plugins we use in our other libs) to fix some violations, like "trailing comma" or "bare except".


jirimoravcik (approved):
nice

Issue #182, opened 2023-08-30
Improve performance of the Python Actor templates


vdusek:
In JS SDK we have [AutoscaledPool](https://docs.apify.com/sdk/js/docs/1.3/api/autoscaled-pool) for executing tasks in parallel. We don't have similar functionality in the Python SDK yet, however, it's planned in the upcoming months.

For now, users could just write some simple utility themselves, using `asyncio.Queue` to get what they need (https://docs.python.org/3/library/asyncio-queue.html#examples).

Writing `AutoscaledPool` might take a while. We could update our templates with some super simple parallelism using `asyncio.Queue`, it might take like 10 extra lines.

The issue is based on the Discord question:

> Hi, I have a custom Python + requests Actor that works great. It's pretty simple, it works against a list of starting URLs and pulls out a piece of information per URL. 
> 
> My question is: If (for example) one run of 1,000 input URLs takes an hour to complete,i would like to parallel-ize it 4 ways so that I can run 4,000 URLs in an hour. 
> 
> What's the best way to do this? I could kick off 4 copies of the run with segmented data, but this seems like something Apify could support natively.
> 
> I saw that if I was using Crawlee (and therefore JS) I could use autoscaling: https://docs.apify.com/platform/actors/running/usage-and-resources . But is there a way to build a single Python based Actor that uses more threads/CPU cores if needed?

vdusek:
I did some initial experiments with the BeautifulSoup Template Actor, comparing the use of the synchronous [requests](https://pypi.org/project/requests/) library to the asynchronous [httpx](https://pypi.org/project/httpx/) library. The Actor code and the input remained the same for both scenarios.

When using `requests`, the execution time was 43 seconds (Actor run - https://console.apify.com/actors/runs/CtmF9T6KRGC5FnJud#log).

When using `httpx`, the execution time improved to 26 seconds (Actor run - https://console.apify.com/actors/runs/7fDkyMe38XThHvIph#log).

Considering that our SDK is already asynchronous-only, there should be no problem in transitioning from the synchronous `requests` to the asynchronous `httpx`. So this will be a first step.

vdusek:
[...]

PR #215, opened 2023-10-05
Use HTTPX instead of Requests in Python templates

vdusek:
### Issue
- https://github.com/apify/actor-templates/issues/182
The motivation for usage of HTTPX is explained here: https://github.com/apify/actor-templates/issues/182#issuecomment-1748615100.

### Description
- All the templates that were using Requests (`python-beautifulsoup`, `python-start`) now use HTTPX.
    - I also updated relevant texts in the readmes and in the code comments. 
- On top of that...
    - I removed `beautifulsoup` from the `requirements.txt` of `python-empty` template (it was not used there).
    - `actor` -> `Actor`
    - Fix a few broken/wrong links (e.g. to JS instead of Python docs)
    - Some unification of `## Included features` section in the README across all the Python Actors

### DIFF
--- a/templates/python-beautifulsoup/.actor/Dockerfile
+++ b/templates/python-beautifulsoup/.actor/Dockerfile
@@ -3,7 +3,7 @@
 # You can also use any other image from Docker Hub.
 FROM apify/actor-python:3.11
 
-# Second, copy just requirements.txt into the actor image,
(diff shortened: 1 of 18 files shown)


jirimoravcik (approved):
nice

PR #219, opened 2023-10-25
chore: Bump Python package versions

fnesveda:
When we've released Apify SDK for Python 1.2.0, the scheduled tests here started failing because the templates didn't include the latest version of the SDK anymore. https://github.com/apify/actor-templates/actions/runs/6633991084/job/18022702321

I've bumped all the Python package versions to fix that. We'll eventually have to figure out some automated way to do this.


fnesveda:
Oops, I accidentally started off from some old branch instead of master, fixed

vdusek (approved):
LGTM

PR #220, opened 2023-10-26
Scrapy: Returning a Request object in the Spider.parse works

### LINKED ISSUE #218: Scrapy: Returning a `Request` object in the `Spider.parse` method does not work
### Problem description

Scrapy users have the option of returning a [scrapy.http.Request](https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request) object from the [scrapy.Spider.parse](https://docs.scrapy.org/en/latest/topics/spiders.html#scrapy.Spider.parse) method. The returned Request will be parsed later. Also, a callback function can be provided to the Request constructor, to specify an alternative function for its processing. 

However, if the Request is returned, it won't be processed in case of executing it on the Apify platform. Because it does not go through the Request Queue. We would have to use [RequestQueue.add_request](https://docs.apify.com/sdk/python/reference/class/RequestQueue#add_request) to prepare a Request for later processing.

### How to replicate it

- Use Python Scrapy Actor template and the following Spider code:
```python
from scrapy import Request, Spider
from scrapy.responsetypes import Response


class TitleSpider(Spider):
    name = 'title'
    start_urls = ['https://www.apc.fr/men/men-shirts.html']

    def parse(self, response: Response) -> Request | Response | None:
        print(f'TitleSpider is parsing {response}...')
        li_elements = response.css('li.product-item')
        product_links = []

        for li in li_elements:
            productlink_container = li.css('.product-link')
            product_link = productlink_container.css('a::attr(href)').get()
            if product_link:
                product_links.append(product_link)

        for link in product_links:
            yield Request(url=link, callback=self.second_callback)

    def second_callback(self, response: Response) -> Request | Response | None:
        print(f'Second callback is parsing {response}...')
        productname = response.css('h1.product-name::text').get()
        print(productname)
```

- Execute the project just using Scrapy, it will work:
```
scrapy crawl title
```

- Execute the project using Apify, it will not work:
```
apify run --purge
```

### Tasks

- Note that more problems were included in this issue, see the comments below.
- [x] Returning of a `Request` object in the `Spider.parse` method works - the `Request` will be scheduled.
- [x] Spider class attribute `start_urls` works.
- [x] All the fields of the Scrapy Request object are stored somewhere (and retrieved later) during the process of Scrapy Request <--> Apify Request transformation.

vdusek:
### Description

- Scrapy: Returning a `Request` object in the `Spider.parse` works.
- Basically, the whole process of mapping between Apify and Scrapy Requests was improved - no information should be lost.

### Issue

- Closes #218 

### Testing

- It was tested with the extended `TitleSpider`, both locally and on the platform.

```python
from typing import Generator, Union
from urllib.parse import urljoin
from scrapy import Request, Spider
from scrapy.responsetypes import Response
from apify import Actor
from ..items import TitleItem

class TitleSpider(Spider):
    """
    Scrapes title pages and enqueues all links found on the page.
    """

    name = 'title_spider'
    start_urls = ['https://apify.com']

    def parse(self, response: Response) -> Generator[Union[TitleItem, Request], None, None]:
        """
        Parse the web page response.

        Args:
            response: The web page response.

        Yields:
            Yields scraped TitleItem and Requests for links.
        """
        Actor.log.info(f'TitleSpider is parsing {response}...')

        # Extract and yield the TitleItem
        url = response.url
        title = response.css('title::text').extract_first()
        yield TitleItem(url=url, title=title, parsed_by='parse')

        # Extract all links from the page, create Requests out of them, and yield them
        for link_href in response.css('a::attr("href")'):
            link_url = urljoin(response.url, link_href.get())
            if link_url.startswith(('http://', 'https://')):
                yield Request(link_url, callback=self.second_parse)

    def second_parse(self, response: Response) -> Generator[Union[TitleItem, Request], None, None]:
        Actor.log.info(f"TitleSpider's second_parse is parsing {response}...")
        # Extract and yield the TitleItem
        url = response.url
        title = response.css('title::text').extract_first()
        yield TitleItem(url=url, title=f'{title}', parsed_by='second_parse')
```


templates/python-scrapy/src/spiders/title.py
```diff
@@ -15,7 +15,7 @@ class TitleSpider(Spider):
     """
 
     name = 'title_spider'
-
+    start_urls = ['https://apify.com']
```
fnesveda, line 18: This will duplicate the `start_urls` from input, is it needed here?

fnesveda (changes requested):
Cool! This looks pretty good 🙂 

The only thing I'm not sure about - do we need to complicate the `TitleSpider` and `TitleItem` with the multiple parses by callbacks? I think it's a pretty advanced feature that is not necessary to have in the template itself, it would just complicate things for novice users.

templates/python-scrapy/src/spiders/title.py
vdusek: Actually, it won't. However, I'll remove it to make the Spider as simple as possible.

templates/python-scrapy/src/items.py
```diff
@@ -12,3 +12,4 @@ class TitleItem(Item):
 
     url = Field()
     title = Field()
+    parsed_by = Field()
```
fnesveda, line 15: Is this still needed with the simpler scraper?

fnesveda (changes requested):
One more nitpick

templates/python-scrapy/src/items.py
vdusek: ofc not, good catch :)

templates/python-scrapy/src/spiders/title.py
```diff
@@ -32,7 +31,7 @@ def parse(self, response: Response) -> Generator[Union[TitleItem, Request], None
         # Extract and yield the TitleItem
         url = response.url
         title = response.css('title::text').extract_first()
-        yield TitleItem(url=url, title=title)
```
fnesveda, line 35: And now the `TitleItem` doesn't have the `parsed_by` parameter 😄

fnesveda (changes requested):
Last one, I promise 😄

templates/python-scrapy/src/spiders/title.py
vdusek: :see_no_evil: ... sorry

fnesveda (approved):
Cool!

PR #223, opened 2023-11-07
feat: add wrapper templates

barjin:
Related to apify/apify-cli#388 and apify/apify-cli#393.

The Scrapy wrapper template is based on, but not a 1:1 copy of the Scrapy Actor template. By maintaining both templates in this repository, we can reuse some of the CI/CD logic and maybe even generate one from the other later on.


src/build_wrappers.js
```diff
@@ -0,0 +1,45 @@
+const fs = require('fs');
+const path = require('path');
+const { execSync } = require('child_process');
+const globby = require('globby');
+const { DIST_DIR_NAME, WRAPPERS_DIR_NAME } = require('./consts');
+
+/**
+ * Creates zips of all wrappers
+ */
+exports.buildWrappers = async function () {
```
fnesveda, line 10: It would be good to have this logic shared somehow across with the `build_templates.js` script.

fnesveda (changes requested):
Cool! I had a few minor suggestions. I was also wondering if we should put this Apify <-> Scrapy logic into a shared package, I proposed that in [Slack](https://apifier.slack.com/archives/C06409LDRM5/p1699376548405789), let's do that there.

vdusek (approved):
LGTM, but I didn't check the JS code, just the wrapper and its structure.

barjin:
@B4nan what do you think? The only things that could be of your interest are the [new manifest](https://github.com/apify/actor-templates/blob/9c59d53a4289e9b9ded40398b66198a4b34e6678/wrappers/manifest.json) for the wrappers (I slimmed that slightly down, opposed to the templates manifest) and the slightly [updated build script](https://github.com/apify/actor-templates/blob/9c59d53a4289e9b9ded40398b66198a4b34e6678/src/build_templates.js), so it builds the wrappers as well.

Merging + releasing this is currently blocking the `apify-cli` PR, so I'd like to have this out asap so we can move on with final round of testing there.

B4nan (approved):

B4nan:
Yeah looking good, sorry for the delay. 

I was also thinking about the package yesterday, and I'd say we should be fine without it as well if the migration is not destructive and only adds new files - we could probably replace old files added by a previous migration with newer versions on a second `apify init --scrapy`, right?

barjin:
Huh, now that's something I didn't think of. I assume it shouldn't cause too many problems - the only thing that could wreak havoc is concatenating the `.gitignore` + `.dockerignore` files... which is quite useful, since without it, git will start tracking the `/storage` folder after the first `apify run`.

Btw, how do we release this two-year-old package? 😬

B4nan:
The templates are downloaded on the fly in the CLI, it's not a package, right? Not sure if they are bundled to the CLI too like we do it in crawlee.

barjin:
> downloaded on the fly in the CLI

They are indeed, but there is a package for fetching the manifest from the correct URL etc. I guess we could just move around this, I just kinda liked the fact that there is no hardcoded URL in the dependent projects.

https://www.npmjs.com/package/@apify/actor-templates

B4nan:
Maybe that's some artifact from the past, and nowadays the templates are always just downloaded, given the package is 2 years old...

B4nan:
If it's needed, we can publish manually.

barjin:
They are fetched manually, but the manifest is still downloaded through this package, see here in [`apify-cli`](
https://github.com/apify/apify-cli/blob/e041cdfda3b8dc50a36ba71c714213435ba4ac54/src/commands/create.js#L50).

If you could publish this with your token, that would be great (I'm not in the maintainers org on NPM).

B4nan:
Right, you changed the implementation of the fetching. Ok will give it a try locally.

B4nan:
Here you go, I hope its enough like this, it only includes the index file as nothing else is referenced from it.

barjin:
seems to be working, cheers 👍🏽

PR #225, opened 2023-11-15
Scrapy-related utilities are moved to the Apify SDK

vdusek:
Closes https://github.com/apify/apify-sdk-python/issues/132


barjin (approved):
Looking good, thanks! :) Did you get a chance to run it with the migrator? It's fine if not - it's not out yet, so we still can check in my PR (and it will be most likely easier).

B4nan (approved):

vdusek:
> It's fine if not - it's not out yet, so we still can check in my PR (and it will be most likely easier).

Was discussed separately. Let's do it this way. Thanks.

PR #236, opened 2023-11-23
Scrapy template: Minor code update and docs improvement

### LINKED ISSUE #235: Python templates: code finetuning, improve docs and comments
We've recently made a minor update to the Scrapy template documentation as part of https://github.com/apify/actor-templates/issues/228.

Let's make further improvements and mainly update the documentation & code comments for other Python templates (BeautifulSoup, Selenium, Playwright, ...) as well.

We should for example let users know, what is safe to change/update/delete and what is not. E.g. during https://github.com/apify/apify-core/issues/13082 we found out that some users just attempted to change the `main` coroutine into a normal function. Which is not possible.

vdusek:
Closes #235

I decided to split solution to the #235 into at least 2 PRs. One for the Scrapy, since it is the biggest template by far, ale the second one will be for other templates.

### DIFF
--- a/wrappers/python-scrapy/{projectFolder}/__main__.template.py
+++ b/wrappers/python-scrapy/{projectFolder}/__main__.template.py
@@ -4,7 +4,7 @@ logging system, and establishing the required environment to run the Scrapy spid
 
 This file is specifically designed to be executed when the project is run as an Apify Actor using `apify run` locally
 or being run on the Apify platform. It is not being executed when running the project as a Scrapy project using
-`scrapy crawl`.
+`scrapy crawl title_spider`.
 
 We recommend you do not modify this file unless you really know what you are doing.
 """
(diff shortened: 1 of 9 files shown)


wrappers/python-scrapy/{projectFolder}/__main__.template.py
```diff
@@ -13,11 +13,15 @@
...
-import scrapy.utils.log
+from scrapy.utils import log as scrapy_logging
+from scrapy.utils.project import get_project_settings
 from apify.log import ActorLogFormatter
 
-# If you want to change the logging level, change it here
-LOGGING_LEVEL = logging.INFO
+# To change the logging level, modify the `LOG_LEVEL` field in `settings.py`. If the field is not present in the file,
+# Scrapy will default to `DEBUG`. This setting applies to all loggers. If you wish to change the logging level for
+# a specific logger, do it in this file.
```
barjin, line 22: Does this mean that for freshly wrapped projects, the log level will always be `DEBUG`?  I assume that a regular `settings.py` in a random Scrapy project doesn't include this field, right?

vdusek: > Does this mean that for freshly wrapped projects, the log level will always be DEBUG?

Not really, the value from the user's `settings.py` will be taken. 

> I assume that a regular `settings.py` in a random Scrapy project doesn't include this field, right?

It depends. `LOG_LEVEL` is a standard settings field - https://docs.scrapy.org/en/latest/topics/settings.html#log-level. So users can set it up. We follow the Scrapy defaults here.

barjin: Ah okay then, no further objections :)

barjin (approved):

PR #239, opened 2023-11-24
Python templates: code finetuning, improve docs and comments

### LINKED ISSUE #235: Python templates: code finetuning, improve docs and comments

vdusek:
Closes #235

### DIFF
--- a/templates/python-beautifulsoup/requirements.txt
+++ b/templates/python-beautifulsoup/requirements.txt
@@ -1,6 +1,7 @@
-# Add your dependencies here.
-# See https://pip.pypa.io/en/latest/reference/requirements-file-format/
-# for how to format them
+# Feel free to add your Python dependencies below. For formatting guidelines, see:
+# https://pip.pypa.io/en/latest/reference/requirements-file-format/
+
 apify ~= 1.3.0
 beautifulsoup4 ~= 4.12.2
-httpx ~= 0.25.0
+httpx ~= 0.25.2
+types-beautifulsoup4 ~= 4.12.0.7
--- a/templates/python-beautifulsoup/src/__main__.py
+++ b/templates/python-beautifulsoup/src/__main__.py
@@ -1,3 +1,10 @@
+"""
+This module serves as the entry point for executing the Apify Actor. It handles the configuration of logging
+settings. The `main()` coroutine is then executed using `asyncio.run()`.
+
+Feel free to modify this file to suit your specific needs.
+"""
+
 import asyncio
 import logging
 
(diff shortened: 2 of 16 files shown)


templates/python-beautifulsoup/requirements.txt
```diff
@@ -1,6 +1,7 @@
-# Add your dependencies here.
-# See https://pip.pypa.io/en/latest/reference/requirements-file-format/
-# for how to format them
+# Feel free to add your Python dependencies below. For formatting guidelines, see:
+# https://pip.pypa.io/en/latest/reference/requirements-file-format/
```
janbuchar, line 2: How hard is it for a template user to transition to poetry, for instance? One day, I'd love to see a link to a guide on that here.

janbuchar (approved):
LGTM

templates/python-beautifulsoup/requirements.txt
vdusek: We might consider offering two versions of the Python templates. One utilizes Poetry and the other uses the default pip with `requirements.txt`.

I would be concerned about updating templates to use Poetry only. Since it is a potential new entry barrier for newcomers unfamiliar with Poetry.

However, we could probably provide 2 versions of the templates, one with Poetry and one with the default pip and `requirements.txt`.

Also, we could integrate the Poetry into SDK & Client & Shared packages for better dependency management. This one would not create any new entry barrier.

I think we've already had some discussions with @fnesveda and @jirimoravcik regarding this topic. Tagging you guys in case you would like to say something about it. Thanks.

fnesveda: We should check if there is demand for that (e.g. if someone did it by themselves). We have some analysis of what packages are users using in their Actors, we could check if someone has an Actor that uses Poetry.

fnesveda: One day I'd love to see some customizable templates, with some wizard to create them, instead of several hard-coded templates.

You could choose:
1. Language (Python, TS, JS)
2. Language version (Python 3.8 - 3.12, Node 18 - 21, or maybe Bun)
3. Technology (Scrapy, Beautiful Soup, Crawlee, Cheerio...)
4. Package manager (NPM, Yarn, Pip, Poetry, ...)
5. Something else?

But that's a dream for the future, I can't imagine the template templating that this would require.

PR #245, opened 2023-11-29
feat(app): comment-out proxy configuration

HonzaTuron:
Closes apify/apify-web#3187

@TheoVasilis Would you check if comment above commented code make sense and if it's clear enough? Thanks!

### DIFF
--- a/templates/js-bootstrap-cheerio-crawler/src/main.js
+++ b/templates/js-bootstrap-cheerio-crawler/src/main.js
@@ -6,10 +6,11 @@ import { router } from './routes.js';
 
 await Actor.init();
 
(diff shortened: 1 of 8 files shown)


TheoVasilis (approved):

jbartadev (approved):

mtrunkat:
@jbartadev @HonzaTuron can we close this? SDK now skips proxy usage when user does not have an access locally. So this is not needed.

B4nan:
Agreed, maybe we should bump the crawlee and SDK versions to ensure the new ones are used (but they should be used anyway, since there is an open version range and no lockfile).

PR #249, opened 2023-12-05
Python Actors: Use `follow_redirects=True` when HTTPX is used

vdusek:
I discovered that HTTPX defaults to `follow_redirects=False`. 

For instance, if a user specifies `https://www.apify.com/` as a starting URL, the resulting output will be as follows:
```json
{
  "url": "https://www.apify.com/",
  "title": "301 Moved Permanently"
}
```

This behavior might not be optimal. In most cases, you would want to follow 301 (or any 3xx) redirections.


barjin (approved):
LGTM, thanks :)

PR #266, opened 2024-01-24
Python Actors: Update to SDK v1.5.5

vdusek:


B4nan (approved):
so it was a bug in the sdk after all :]

PR #273, opened 2024-03-13
chore: update version requirements for python sdk

janbuchar:
This should fix the failing tests. We might want to make the requirements looser and just install the latest version.


B4nan (approved):
i dont recall why the version was locked to ~1.6, maybe using ~1 would be enough?

janbuchar:
> i dont recall why the version was locked to ~1.6, maybe using ~1 would be enough?

I guess we can wait for @vdusek and revisit this later.

PR #277, opened 2024-06-13
fix: python playwright template

vdusek:
based on [Slack bug report](https://apify.slack.com/archives/C0L33UM7Z/p1718270319985559):

```text
There is a problem with the [Playwright + Chrome template](https://console.apify.com/actors/templates/python-playwright). When you use the template, build the Actor and try to run it, it returns this to the log:

playwright._impl._api_types.Error: Executable doesn't exist at /root/.cache/ms-playwright/chromium-1084/chrome-linux/chrome
Looks like Playwright was just installed or updated.
Please run the following command to download new browsers:
playwright install

And then, when you add RUN playwright install to the Dockerfile, there is one more error, this time with playwright install-deps.

Only then, after you add RUN playwright install-deps to the Dockerfile, it finally works.
```

### DIFF
--- a/templates/python-playwright/.actor/Dockerfile
+++ b/templates/python-playwright/.actor/Dockerfile
@@ -20,6 +20,10 @@ RUN echo "Python version:" \
  && echo "All installed Python packages:" \
  && pip freeze
 
+ # Install Playwright and its dependencies
+RUN playwright install-deps && \
+    playwright install
+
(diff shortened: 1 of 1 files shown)


janbuchar (approved):
The problem may be caused by a playwright version mismatch between the base image and this. Anyway, the changes can't hurt it - let's do this.

PR #282, opened 2024-08-06
Update ts-start-bun template prefill url property

MFori:
Template will use same prefill demo url as other templates - crawlee.dev without www subdomain as the www variant is failing on 404


B4nan (approved):
please ignore the broken tests, we are aware of that, they broke with the recent CLI release

valekjo (approved):

PR #283, opened 2024-08-12
fix: incorrect input_schema.json titles

DaveHanns:
Fix incorrect input_schema.json titles in crawlee-puppeteer-chrome templates


jirimoravcik (approved):

B4nan (approved):
good catch, heh

gippy (approved):

PR #288, opened 2024-09-03
fix: python templates use SDKv2 and Crawlee templates are added

vdusek:
Merging is blocked until SDK v2 is released.

Also do not forget to revert tmp changes in `test/templates.test.js`.


B4nan:
the scrapy template test is now skipped here

https://github.com/apify/actor-templates/blob/master/templates/manifest.json#L13

janbuchar (approved):
I assume the SDK version will be set to ~=2.0.0 when 2.0 is out :slightly_smiling_face:

templates/python-scrapy/src/__main__.py
```diff
@@ -9,6 +10,8 @@
 We recommend you do not modify this file unless you really know what you are doing.
 """
 
+# TODO: explicitne vypneme apify logging configuration v actorovi, zde to udelame manualne
```
janbuchar, line 13: Huh? :slightly_smiling_face:

templates/python-selenium/src/main.py
```diff
@@ -13,77 +12,91 @@
...
-            depth = request['userData']['depth']
+        # Process the URLs from the request queue.
+        while request := await request_queue.fetch_next_request():
+            url = request.url
+            depth = request.user_data['depth']
             Actor.log.info(f'Scraping {url} ...')
 
             try:
-                # Open the URL in the Selenium WebDriver
+                # Navigate to the URL using Selenium WebDriver.
```
janbuchar, line 72: It probably won't help much with anything, but maybe we could wrap `driver.get` with `asyncio.to_thread` here?

janbuchar (approved):
I can't say I read all of this carefully, but I tried. LGTM.

PR #292, opened 2024-09-06
fix: Add 60 minutes timeout to template tests

fnesveda:
The broken template tests cost us $500 in the last month, because the tests were running for 6 hours before they timed out. This adds a 1-hour time limit to them, so that next time they time out earlier and it's not so costly.

I checked and the tests typically run for 30 minutes, so 1 hour timeout should be enough.

### DIFF
--- a/.github/workflows/lint_and_test.yaml
+++ b/.github/workflows/lint_and_test.yaml
@@ -54,6 +54,7 @@ jobs:
         node-version: [ 18, 20, 22 ]
       fail-fast: false
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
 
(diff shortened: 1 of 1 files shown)


B4nan:
Wait a sec, this is public repository, I thought we don't pay for the GH actions here?

B4nan (approved):
either way it's a good idea. if we really pay for this i feel dumb for not prioritizing the fix (not that we didn't try several times already in the past)

fnesveda:
Ah, fair, I didn't realize GitHub Actions are free for public repositories 🤦 I checked our organization billing and I think it should be free.

PR #293, opened 2024-09-12
docs: update titles of python templates

vdusek:
it should be (more) in accordance with the JS templates

### DIFF
--- a/templates/manifest.json
+++ b/templates/manifest.json
@@ -66,7 +66,7 @@
         {
             "id": "python-beautifulsoup",
             "name": "python-beautifulsoup",
-            "label": "BeautifulSoup + HTTPX",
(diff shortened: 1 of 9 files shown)


janbuchar (approved):

templates/manifest.json
```diff
@@ -219,7 +219,7 @@
         {
             "id": "python-crawlee-playwright",
             "name": "python-crawlee-playwright",
-            "label": "Start with Python Crawlee and Playwright",
+            "label": "Crawlee + Playwright + Chrome",
```
B4nan, line 222: but the template is using chromium now, right? 🙃

janbuchar: Correct. Feel free to open an issue to straighten that up :slightly_smiling_face:

PR #294, opened 2024-09-16
fix: Add configs to ts templates

MatousMarik:
- added ".vscode" to `.gitignore` and `.dockerignore` files
- added `.editorconfig` files to templates if missing
- added or updated `.eslintrc`
- updated `package.json` with eslint packages
- removed duplicate option in `ts-bootstrap-cheerio-crawler/tsconfig.json`
- non-null assertion fix in `main.ts` in _ts-start_ and _ts-start-bun_ templates

### DIFF
--- a/templates/ts-bootstrap-cheerio-crawler/.eslintrc
+++ b/templates/ts-bootstrap-cheerio-crawler/.eslintrc
@@ -1,4 +1,20 @@
 {
     "root": true,
-    "extends": "@apify/eslint-config-ts"
+    "env": {
+        "browser": true,
+        "es2020": true,
+        "node": true
+    },
+    "extends": [
+        "@apify/eslint-config-ts"
+    ],
+    "parserOptions": {
+        "project": "./tsconfig.json",
+        "ecmaVersion": 2020
+    },
+    "ignorePatterns": [
+        "node_modules",
+        "dist",
+        "**/*.d.ts"
+    ]
 }
--- a/templates/ts-empty/package.json
+++ b/templates/ts-empty/package.json
@@ -11,7 +11,11 @@
         "crawlee": "^3.5.4"
     },
     "devDependencies": {
+        "@apify/eslint-config-ts": "^0.3.0",
         "@apify/tsconfig": "^0.1.0",
+        "@typescript-eslint/eslint-plugin": "^6.7.2",
+        "@typescript-eslint/parser": "^6.7.2",
+        "eslint": "^8.50.0",
         "tsx": "^4.6.2",
         "typescript": "^5.3.3"
     },
--- a/templates/ts-bootstrap-cheerio-crawler/.dockerignore
+++ b/templates/ts-bootstrap-cheerio-crawler/.dockerignore
@@ -1,5 +1,6 @@
 # configurations
 .idea
+.vscode
 
 # crawlee and apify storage folders
 storage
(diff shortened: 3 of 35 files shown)


templates/ts-bootstrap-cheerio-crawler/.eslintrc
```diff
@@ -1,4 +1,20 @@
...
+    ],
+    "parserOptions": {
+        "project": "./tsconfig.json",
+        "ecmaVersion": 2020
+    },
+    "ignorePatterns": [
+        "node_modules",
+        "dist",
+        "**/*.d.ts"
+    ]
```
B4nan, line 19: do you really need those changes? you shouldn't need most of them i think. and if you need some, it might be better to modify the shared configs.

https://github.com/apify/apify-eslint-config/blob/master/index.js
https://github.com/apify/apify-eslint-config-ts/blob/master/index.js

looking at them, we already set some of those options.

templates/ts-empty/package.json
```diff
@@ -11,7 +11,11 @@
         "crawlee": "^3.5.4"
     },
     "devDependencies": {
+        "@apify/eslint-config-ts": "^0.3.0",
         "@apify/tsconfig": "^0.1.0",
+        "@typescript-eslint/eslint-plugin": "^6.7.2",
+        "@typescript-eslint/parser": "^6.7.2",
```
B4nan, line 17: why not v7?

B4nan:
hmm also the template tests apparently wont run if you PR from fork, i will invite you to the project

templates/ts-empty/package.json
MatousMarik: I found this version in other templates, and I don't know where to find the actual used version across projects (the latest version for @typescript-eslint/parser is 8.5)

templates/ts-bootstrap-cheerio-crawler/.eslintrc
MatousMarik: In this case, I copy pasted .eslintrc from another template (it is the same in all other templates)
Shortcuts:
- *a-config* = https://github.com/apify/apify-eslint-config/blob/master/index.js
- *a-config-ts* = https://github.com/apify/apify-eslint-config-ts/blob/master/index.js

As I see it, necessary would be only:
- env: es2020: true
  - or it could be added to a-config
- extends: @apify/eslint-config-ts
- parserOptions: ecmaVersion: 2020 
  - ?? I am not sure about this one - it is in a-config, but iparserOptions are under "overrides" in a-config-ts without ecmaVersion: 2020 so probably necessary here, but could be added to a-config-ts
- ignorePatterns

Do you think I should change it?
For the record, I joined Apify two weeks ago and don't want to make any significant changes.

templates/ts-empty/package.json
B4nan: > the latest version for @typescript-eslint/parser is 8.5

yes, and latest version of eslint is v9, we cant use those now since they require the new flat configs. but we can use v7 just fine.

MatousMarik: Should I change it and make a new PR from the branch from the upstream repo?
Check if I understand you correctly:
-> "@typescript-eslint/eslint-plugin": "^7.18.0"
-> "@typescript-eslint/parser": "^7.18.0"

B4nan: yes please, otherwise the tests wont run. latest v7 should be fine, we use that in all our repos afaik.

PR #297, opened 2024-10-22
chore: bump templates to latest crawlee & apify version

metalwarrior665:
This is done manually like this once a while?


B4nan:
There are no lockfiles, we dont need to do this at all, installing the template will use the latest versions of both.

metalwarrior665:
Yeah sure, it is just about how it looks

PR #307, opened 2024-12-18
chore: bump `ts-playwright-test-runner` Playwright and Node version

barjin:
Bumps the Playwright and Node versions in the template to `latest`s. Fixes the hanging `browserContext.newPage()` call by passing the `--disable-gpu` launch option to the browser.

More context at https://apify.slack.com/archives/CGZSN9DQC/p1733928283797729

### DIFF
--- a/templates/js-crawlee-playwright-chrome/src/main.js
+++ b/templates/js-crawlee-playwright-chrome/src/main.js
@@ -24,6 +24,13 @@ const proxyConfiguration = await Actor.createProxyConfiguration();
 const crawler = new PlaywrightCrawler({
     proxyConfiguration,
     requestHandler: router,
+    launchContext: {
+        launchOptions: {
+            args: [
+                '--disable-gpu', // Mitigates the "crashing GPU process" issue in Docker containers
+            ]
+        }
+    }
 });
 
 await crawler.run(startUrls);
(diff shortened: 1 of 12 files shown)


barjin:
I see that other templates using Playwright (e.g. [ts-playwright-chrome](https://github.com/apify/actor-templates/blob/master/templates/ts-crawlee-playwright-chrome/.actor/Dockerfile) or [js-playwright-chrome](https://github.com/apify/actor-templates/blob/master/templates/js-crawlee-playwright-chrome/.actor/Dockerfile)) use the unpinned base Docker image (`apify/actor-node-playwright-chrome:20`, without the playwright version tag).

What's the consensus on this? I remember that there were some nasty bugs caused by the installed browser / npm-installed Playwright version mismatch, right? 

Should we unpin Playwright here too? Or should we rather pin the Playwright version in the other templates?

B4nan:
> What's the consensus on this? I remember that there were some nasty bugs caused by the installed browser / npm-installed Playwright version mismatch, right?

My point of view is that we should pin it, but we are missing some automation to keep the templates up to date, I am not sure if renovate can pick this up properly. And the docs of course, I am not sure if we have some suggestions on the "old way" (don't commit lockfile, use `*` in deps and don't pin in docker tag).

> And, on related note, shouldn't we pass the --disable-gpu flag to the other templates as well? Since those are "Apify templates", I'd assume they work perfectly on Apify, no user tweaks ever needed.

Makes sense to me.

B4nan (approved):

PR #308, opened 2025-01-13
fix: Fix Python templates to work with Crawlee v0.5

vdusek:
- Pin Crawlee & Scrapy versions
- Fixing Scrapy template


Pijukatel (approved):
One of the issues found during investigation of this,  was delegated to another issue:
https://github.com/apify/actor-templates/issues/309
That can be solved later.

PR #315, opened 2025-02-17
feat: add llamaindex agent template

jirispilka:
@vdusek We're preparing templates for AI Agents to be ready for the Thursday (Feb, 17) hackathon. I’d really appreciate it if you or @Pijukatel could review them.


jirispilka:
I have not touched nodejs templates, `puppeteer` tests were failing before

templates/python-llamaindex-agent/.actor/actor.json
```diff
@@ -0,0 +1,13 @@
+{
+    "actorSpecification": 1,
+    "name": "python-llamaindex-agent",
+    "title": "Pyton LlamaIndex Agent",
```
MQ37, line 4:
```suggestion
    "title": "Python LlamaIndex Agent",
```

templates/python-llamaindex-agent/.actor/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+ && pip freeze
+
+# Next, copy the remaining files and directories with the source code.
+# Since we do this after installing the dependencies, quick build will be really fast
+# for most source file changes.
+COPY . ./
+
+# Use compileall to ensure the runnability of the Actor Python code.
+RUN python3 -m compileall -q .
+
```
MQ37, line 30:
```suggestion

# Create and run as a non-root user.
RUN useradd -m apify && \
    chown -R apify:apify ./
USER apify

```

MQ37 (approved):
Found typo and added non-root user to docker container, otherwise LGTM  :+1:

vdusek (changes requested):
Nice! I have a few comments...

On top of that, I realized we don't have static analysis for Python templates 👀. I've added it here: [#316](https://github.com/apify/actor-templates/pull/316) - feel free to check out that PR as well.

Once it's merged, it'd be great if you could rebase on top of it so this Actor is also linted and type-checked.

On top of that, I realized we do not have a static analysis for Python templates :eyes:. I am adding it here - https://github.com/apify/actor-templates/pull/316 - you can check that PR as well. Also would be great once it is merged you rebased upon that to have these Actor linted & type checked as well.

templates/python-llamaindex-agent/.dockerignore
Pijukatel, line 1: (Same comment for .gitignore)
This will do it's job, but it might be overkill for this project as many of those ignore rules are out of scope of this. Maybe delete at least rules, we are sure are not needed in this context.

templates/python-llamaindex-agent/README.md
```diff
@@ -0,0 +1,45 @@
...
+- **[Apify SDK](https://docs.apify.com/sdk/python/)** for Python - a toolkit for building Apify [Actors](https://apify.com/actors) and scrapers in Python.
+- **[Input Schema](https://docs.apify.com/platform/actors/development/input-schema)** - define and easily validate a schema for your Actor's input.
+- **[LlamaIndex](https://github.com/run-llama/llama_index)** - a framework for building LLM-powered agents using your data.
+- **[Dataset](https://docs.apify.com/sdk/python/docs/concepts/storages#working-with-datasets)** - a storage solution for structured data where each object stored shares the same attributes.
+
+## How it works
+
+The Agent has two main tools:
+
+1. `call_contact_details_scraper` - Calls the [Contact Details Scraper](https://apify.com/apify/contact-details-scraper) to scrape contact details from websites.
```
Pijukatel, line 17: Seems to me this link is not correct at the moment.
I found it here instead:
https://apify.com/vdrmota/contact-info-scraper

templates/python-llamaindex-agent/src/agent.py
```diff
@@ -0,0 +1,44 @@
+import logging
+
+from llama_index.core.agent import ReActAgent
+from llama_index.core.chat_engine.types import AgentChatResponse
+from llama_index.core.tools import FunctionTool
+from llama_index.llms.openai import OpenAI
+
+from .tools import LLMRegistry, call_contact_details_scraper, summarize_contact_information
+
+tool_call_contact_details_scraper = FunctionTool.from_defaults(fn=call_contact_details_scraper)
```
Pijukatel, line 10: Does this have to be done during import time? I haven't see this being imported in other files, maybe it can be done directly on line 35

templates/python-llamaindex-agent/src/main.py
```diff
@@ -0,0 +1,71 @@
...
+"""
+
+import os
+
+from apify import Actor
+from llama_index.llms.openai import OpenAI
+
+from .agent import run_agent
+
+fallback_input = {
```
Pijukatel, line 19: This is for local testing and should be deleted when running on Platform?

jirispilka:
Thank you for the review!  

@vdusek, I was actually wondering the same—there was no linting or checks for Python templates.  

Thanks for adding it so quickly! As you can see, there were quite a few changes, mostly because I used double quotes instead of single quotes 🙈.  But a couple of others issues were actually pretty serious 😊 

I've addressed most of the issues, except for a couple from @Pijukatel  regarding `.gitignore` and Pydantic.  

I agree with @Pijukatel  that `fallback_input` is non-standard and a bit hacky, but I haven’t figured out a better way to pass CI.

vdusek (approved):
LGTM

templates/python-llamaindex-agent/src/tools.py
```diff
@@ -1,52 +1,60 @@
...
         cls._llm = new_llm
 
 
 async def call_contact_details_scraper(
     start_urls: list[dict[str, Any]],
     max_requests_per_start_url: int = 20,
     max_depth: int = 2,
-    same_domain: bool = True,
-    deduplicate: bool = True,
+    same_domain: bool = True,  # noqa:FBT001,FBT002
```
Pijukatel, line 54: I think you can get rid of the errors if you use `*` to enforce keyword arguments at least for those last two bool kwargs. It makes calls to this method more readable
`call_contact_details_scraper(20,2,True,True)`
vs
`call_contact_details_scraper(20,2,same_domain=True,deduplicate=True)`

Pijukatel (approved):

PR #316, opened 2025-02-18
chore: add static analysis for Python templates

vdusek:
Add static analysis (Ruff, Mypy) for Python templates.

Edit: Mypy will require further updates due to our usage of `src/`:

```
$ make type-check 
poetry run mypy
templates/python-crawlee-beautifulsoup/src/__init__.py: error: Duplicate module named "src" (also at "templates/python-beautifulsoup/src/__init__.py")
templates/python-crawlee-beautifulsoup/src/__init__.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
templates/python-crawlee-beautifulsoup/src/__init__.py: note: Common resolutions include: a) using `--exclude` to avoid checking one of them, b) adding `__init__.py` somewhere, c) using `--explicit-package-bases` or adjusting MYPYPATH
Found 1 error in 1 file (errors prevented further checking)
make: *** [Makefile:16: type-check] Error 2
```

This will be addressed later in [#317](https://github.com/apify/actor-templates/pull/317#317).


.gitignore
```diff
@@ -17,6 +17,6 @@ __pycache__
 .mypy_cache
 .pytest_cache
 .ruff_cache
-pyproject.toml
 .venv
 venv
+poetry.toml
```
Pijukatel, line 22: Why is this added. Do you use it for some local overrides?

vdusek: I do, as I prefer to have the `.venv` inside the project directory rather than in some central location.

Pijukatel (approved):

PR #318, opened 2025-02-18
feat: add LangGraph.js Agent into templates

jirispilka:


templates/js-langgraph-agent/src/main.js
```diff
@@ -0,0 +1,97 @@
...
+//    https://platform.openai.com/account/api-keys. Refer to
+//    https://docs.apify.com/cli/docs/vars#set-up-environment-variables-in-apify-console for guidance
+//    on setting environment variables.
+const { OPENAI_API_KEY, APIFY_TOKEN } = process.env;
+
+// You can configure the input for the Actor in the Apify UI when running on the Apify platform or editing
+// storage/key_value_stores/default/INPUT.json when running locally.
+const {
+    // query = 'How to build LangGraph agent at Apify platform?',
+    query = 'This is fallback test query, do nothing and ignore it.',
```
jirimoravcik, line 40: Please clean up the comment.
If there's no query provided, wouldn't it be better to crash? It doesn't make sense to run the code without a valid query.

jirimoravcik (approved):
Looks good.
The only thing I'm concerned about is that people prefer TypesScript

jirispilka:
> Looks good. The only thing I'm concerned about is that people prefer TypesScript

hmm, I should have started with TypeScript. I might do it later

PR #319, opened 2025-02-19
feat: python-crewai template

MQ37:


templates/python-crewai/src/ppe_utils.py
```diff
@@ -0,0 +1,37 @@
...
+    Actor.log.debug(f'Charging for {tokens_hundreds} hundred tokens for model {model_name}')
+
+    if model_name == 'gpt-4o':
+        await Actor.charge(event_name='openai-100-tokens-gpt-4o', count=tokens_hundreds)
+    elif model_name == 'gpt-4o-mini':
+        await Actor.charge(event_name='openai-100-tokens-gpt-4o-mini', count=tokens_hundreds)
+    elif model_name == 'o1':
+        await Actor.charge(event_name='openai-100-tokens-o1', count=tokens_hundreds)
+    elif model_name == 'o3-mini':
+        await Actor.charge(event_name='openai-100-tokens-o3-mini', count=tokens_hundreds)
```
jirimoravcik, line 26: create a dict mapping instead of copy-paseted `Actor.charge` please

templates/python-crewai/src/tools.py
```diff
@@ -0,0 +1,83 @@
...
+    Raises:
+        RuntimeError: If the Actor fails to start.
+    """
+    run_input = {
+        'directUrls': [f'https://www.instagram.com/{handle}/'],
+        'resultsLimit': max_posts,
+        'resultsType': 'posts',
+        'searchLimit': 1,
+    }
+    apify_client = ApifyClient(token=Actor.config.token)
```
jirimoravcik, line 51: I don't like the `Actor.config.token` usage, can this be done differently?

templates/python-crewai/src/main.py
```diff
@@ -0,0 +1,104 @@
...
+    the field of web scraping significantly.
+
+    Raises:
+        ValueError: If the input is missing required attributes.
+    """
+    async with Actor:
+        # Handle input
+        actor_input = await Actor.get_input()
+        # fallback input is provided only for testing, you need to delete this line
+        actor_input = {**fallback_input, **actor_input}
```
jirimoravcik, line 40: Again, I don't like this, it shouldn't be here at all. You should always test it with some `INPUT.json` file.

jirimoravcik (changes requested):
Thanks, had some small points

templates/python-crewai/src/ppe_utils.py
MQ37: agree, forgot to refactor the initial implementation where each model used different number of tokens based on openai pricing, thank you for noticing :+1:

templates/python-crewai/src/tools.py
MQ37: I needed an Apify sync client and this approach, getting token from Actor config, is also employed by the Actor SDK internal Apify client (that is async only unfortunately) so I thought this is the way. Is user's API token stored somewhere else that is accessible from the Actor run?

jirispilka: Env variable `APIFY_TOKEN` https://docs.apify.com/platform/actors/development/programming-interface/environment-variables

jirispilka (approved):
Let's create an issue with comments from @jirimoravcik , link it in PR and merge.
We will fix it after or during the hackathon

jirimoravcik (approved):

MQ37:
Created issue https://github.com/apify/actor-templates/issues/325 will address suggestions and tests after the hackathon

PR #324, opened 2025-02-19
feat: add llama index ppe

jirispilka:


MQ37 (approved):
Did  extremely quick pass. In my PRs I am charging per token, but this is also a way. LGTM.

jirispilka:
> Did extremely quick pass. In my PRs I am charging per token, but this is also a way. LGTM.

Yeah, I wanted to do it quickly :)

PR #327, opened 2025-02-20
fix: add package-lock.json back

MQ37:


B4nan:
FYI templates shouldn't have lockfiles, we want users to have latest (in range) dependencies when they create the project from a template

MQ37:
> FYI templates shouldn't have lockfiles, we want users to have latest (in range) dependencies when they create the project from a template
> 
> edit: hmm or is that a root lockfile? that would be fine

Hey, yes I accidentally removed the root package lock :grin:

Issue #325, opened 2025-02-20
Address review suggestions and fix tests for python-crewai template 


MQ37:
https://github.com/apify/actor-templates/pull/319

**To address**:

- [x] https://github.com/apify/actor-templates/pull/319#discussion_r1961944370
- [ ] https://github.com/apify/actor-templates/pull/319#discussion_r1961948066

MQ37:
merged https://github.com/apify/actor-templates/pull/328

PR #328, opened 2025-02-20
fix: python-crewai template

MQ37:
https://github.com/apify/actor-templates/issues/325


MQ37:
Python tests for 3.9 will need to skip the `python-crewai` template since it is available for Python >= 3.10

B4nan:
please prioritize the fix for the broken CI

MQ37:
3.13 also needs to be skipped - crewai is >=3.10 and < 3.13

jirimoravcik (approved):

PR #332, opened 2025-02-25
fix: docker non root

MQ37:
**Changes**:
- added non-root docker user to majority of templates (except playwright ones)
- added tests for template docker image build (js-standby example was not building)


jirispilka (changes requested):
Should we have the same command for creating user in all dockerfiles?
```
RUN adduser -D -H apify && \
    chown -R apify:apify ./
USER apify
```
vs 
````
# Create and run as a non-root user.
RUN useradd --no-create-home apify && \
    chown -R apify:apify ./
USER apify
````

I don’t have much experience with writing GitHub workflows, but embedding a script directly inside the YAML file seems suspicious. It could make maintenance and debugging more difficult. 

I leave this to @jirimoravcik and @vdusek to judge it.

File: .github/workflows/test_dockerfile_builds.yml [hunk 1]
```diff
@@ -0,0 +1,77 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Test Dockerfile builds
+        shell: bash
+        run: |
+          mapfile -t dockerfiles < <(find templates -name "Dockerfile" -type f)
```
jirimoravcik, line 24: Could you please move this into a separate `.sh` file? Thanks

jirimoravcik (changes requested):
Thanks.
I'd unify the `adduser` commands as I see some discrepancies between them in files.

MQ37:
> Should we have the same command for creating user in all dockerfiles?
> 
> ```
> RUN adduser -D -H apify && \
>     chown -R apify:apify ./
> USER apify
> ```
> 
> vs
> 
> ```
> # Create and run as a non-root user.
> RUN useradd --no-create-home apify && \
>     chown -R apify:apify ./
> USER apify
> ```
> 
> I don’t have much experience with writing GitHub workflows, but embedding a script directly inside the YAML file seems suspicious. It could make maintenance and debugging more difficult.
> 
> I leave this to @jirimoravcik and @vdusek to judge it.

They are not unified because some imags use different base OS image and some templates require user home and some don't. WIll unify them so they all create home and use same options. @jirimoravcik @jirispilka

.github/workflows/test_dockerfile_builds.yml
MQ37: Sure, will split into separate file :+1:

jirimoravcik (approved):

MQ37:
> Could you please add to the description something about "x-apify-container-server-readiness-probe". I didn't understand how is it related to the rest of the change.

Sure, added a comment regarding the readiness probe. Thank you for suggestion :+1:

Pijukatel (approved):
LGTM

jirispilka (approved):

PR #334, opened 2025-02-26
feat: Add CLI template

janbuchar:
This might come in handy for people who want to try Actorizing existing code with the Apify CLI

### DIFF
--- a/templates/cli-start/.actor/Dockerfile
+++ b/templates/cli-start/.actor/Dockerfile
@@ -1,0 +1,26 @@
+# Specify the base Docker image. You can read more about
+# the available images at https://docs.apify.com/sdk/js/docs/guides/docker-images
+# You can also use any other image from Docker Hub.
+FROM apify/actor-node:22
+
+# Install system dependencies
+RUN apk add --no-cache jq wget bash
+
+# Install the Apify CLI
+RUN npm --quiet set progress=false \
+    && npm install -g apify-cli \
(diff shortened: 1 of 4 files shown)


templates/cli-start/.actor/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+FROM apify/actor-node:22
+
+# Install system dependencies
+RUN apk add --no-cache jq wget bash
+
+# Install the Apify CLI
+RUN npm --quiet set progress=false \
+    && npm install -g apify-cli \
+    && echo "Installed NPM packages:" \
+    && (npm list --omit=dev --all || true) \
```
B4nan, line 13: is this doing something for globally installed packages?

templates/cli-start/.actor/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+    && echo "Node.js version:" \
+    && node --version \
+    && echo "NPM version:" \
+    && npm --version \
+    && rm -r ~/.npm
```
B4nan, line 18: don't you need `-f` too for folders?

B4nan (approved):

templates/cli-start/.actor/Dockerfile
janbuchar: Huh. Truth be told, this is just copypasted. I doubt it'd work on nonempty folders, so I might as well just remove it.

PR #336, opened 2025-02-26
feat: add Camoufox-based Crawlee JS/TS templates

barjin:
Following the https://github.com/apify/apify-sdk-js/pull/364 and https://github.com/apify/crawlee/pull/2842 , this PR adds Camoufox-enabled templates to Apify Actor templates. The implementation is heavily based on the existing Playwright + Chrome templates.

The only issue (I'm aware of) currently is the immense size of those images (as they contain Chrome and we add Camoufox binaries). Installing Camoufox directly to a `node-debian` image results in missing system dependencies. While it might be possible to install those manually in the Dockerfile, it might make the Dockerfile too complex for a regular user.

image

### DIFF
--- a/templates/js-crawlee-playwright-camoufox/README.md
+++ b/templates/js-crawlee-playwright-camoufox/README.md
@@ -1,0 +1,21 @@
+## PlaywrightCrawler template
+
+This template is a production-ready boilerplate for developing an [Actor](https://apify.com/actors) with `PlaywrightCrawler`. Use this to bootstrap your projects using the most up-to-date code.
+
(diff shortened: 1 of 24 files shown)


templates/js-crawlee-playwright-camoufox/README.md
```diff
@@ -0,0 +1,21 @@
+## PlaywrightCrawler template
```
B4nan, line 1: this should mention camoufox, at least in the heading

templates/ts-crawlee-playwright-camoufox/README.md
```diff
@@ -0,0 +1,21 @@
+## PlaywrightCrawler template
```
B4nan, line 1: same here

B4nan (approved):
looks good, i would just update the readmes to mention camoufox

PR #339, opened 2025-03-05
fix: update crewai template

MQ37:
Clean code, token based ppe -> task based, align with the proposed build an AI agent article


jirispilka (approved):
Awesome. Thanks!

PR #342, opened 2025-03-06
fix: python-crawlee-playwright add no sandbox

MQ37:
based on https://apify.slack.com/archives/C0L33UM7Z/p1741278934497689


B4nan (approved):

B4nan:
the broken CI will be fixed via #345

PR #343, opened 2025-03-07
fix: unify agent ppe events

MQ37:


jirispilka (approved):
Looks good to me, I would only suggest unifying the text too:
```
Price for actor start
Price for completing the task
```

PR #348, opened 2025-03-10
chore: update Python Crawlee versions

vdusek:


Pijukatel (approved):

templates/python-crawlee-beautifulsoup/requirements.txt
```diff
@@ -2,4 +2,4 @@
 # https://pip.pypa.io/en/latest/reference/requirements-file-format/
 
 apify < 3.0
-crawlee[beautifulsoup] < 0.6
```
janbuchar, line 5: Just a thought - couldn't we use `*` instead? The SDK should limit it to the newest compatible version already

vdusek: Makes sense, thanks.

janbuchar (approved):

Issue #363, opened 2025-03-28
Adjust default URls in input schemas, default inputs and route-handlers


JJetmar:
I've noticed that some templates (`templates/ts-crawlee-playwright-chrome`, `templates/ts-crawlee-puppeteer-chrome`) includes irregularities within default urls in input schema, url in default input and pattern for enqueuing new requests in the request-handlers.

If user runs templates with any changes it leads to an empty dataset. No Requests are being enqueued as default input is `https://crawlee.dev` but queuing next requests require pattern `https://apify.com/*` and etc.

PR #364, opened 2025-03-28
fix: Adjust default URls in input schemas, default inputs and route-handlers (#363)

JJetmar:
Related to #363

### DIFF
--- a/templates/ts-crawlee-playwright-chrome/src/routes.ts
+++ b/templates/ts-crawlee-playwright-chrome/src/routes.ts
@@ -5,7 +5,7 @@ export const router = createPlaywrightRouter();
 router.addDefaultHandler(async ({ enqueueLinks, log }) => {
     log.info(`enqueueing new URLs`);
     await enqueueLinks({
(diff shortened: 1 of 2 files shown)


Pijukatel (approved):

B4nan (commented):
can we rather unify what the JS and TS templates are scraping?

Pijukatel:
I saw that fork based PR did not trigger the tests, so I will add this trigger there.

B4nan:
both options are fine with me, if we can share secrets based on explicit approval, it's all right. and yes, ideally people from engineering, including delivery, should just PR from branch instead of a fork ideally.

JJetmar:
> can we rather unify what the JS and TS templates are scraping?
@B4nan This PR solves Issue with zero results in the default templates. Are you OK with me merging it? You can create separate Issue for adjusting what websites are being scraped.

B4nan:
I will PR this myself, all I am asking is the same 2 loc PR, just done differently.

Issue #358, opened 2025-03-20
Add type checking for Python AI templates


vdusek:
- Add type checking for Python AI templates:
  - `python-crewai`,
  - `python-langgraph`,
  - `python-llamaindex-agent`.
- See: [Makefile#L15:L26](https://github.com/apify/actor-templates/blob/9d572dc1d4e3a5e886afc60c187356bd3ca9c26c/Makefile#L15:L26).

PR #365, opened 2025-03-28
fix: add type checks for ai py templates

MQ37:
closes https://github.com/apify/actor-templates/issues/358

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,3 +87,4 @@ warn_redundant_casts = true
 warn_return_any = true
 warn_unreachable = true
 warn_unused_ignores = true
+ignore_missing_imports = true
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,9 @@ type-check:
 	uv run mypy templates/python-selenium
 	uv run mypy templates/python-standby
 	uv run mypy templates/python-start
+	uv run mypy templates/python-crewai
+	uv run mypy templates/python-langgraph
+	uv run mypy templates/python-llamaindex-agent
 
 format:
 	uv run ruff check --fix
--- a/templates/python-langgraph/src/tools.py
+++ b/templates/python-langgraph/src/tools.py
@@ -15,7 +15,7 @@ from langchain_core.tools import tool
 from src.models import InstagramPost
 
 
-@tool
+@tool # type: ignore
 def tool_calculator_sum(numbers: list[int]) -> int:
     """Tool to calculate the sum of a list of numbers.
 
@@ -28,7 +28,7 @@ def tool_calculator_sum(numbers: list[int]) -> int:
     return sum(numbers)
 
 
-@tool
+@tool # type: ignore
 async def tool_scrape_instagram_profile_posts(handle: str, max_posts: int = 30) -> list[InstagramPost]:
     """Tool to scrape Instagram profile posts.
 
--- a/templates/python-llamaindex-agent/src/main.py
+++ b/templates/python-llamaindex-agent/src/main.py
@@ -65,3 +65,4 @@ async def run_query(query: str, model_name: str) -> AgentChatResponse | None:
     except Exception as e:
         msg = f'Error running LlamaIndex Agent, error: {e}'
         await Actor.fail(status_message=msg, exception=e)
+    return None
--- a/templates/python-llamaindex-agent/src/tools.py
+++ b/templates/python-llamaindex-agent/src/tools.py
@@ -99,7 +99,7 @@ async def call_contact_details_scraper(
     }
     logger.info(f'Calling Apify Actor: {CONTACT_DETAILS_ACTOR_ID} with input: {run_input}')
     actor_call = await client.actor(CONTACT_DETAILS_ACTOR_ID).call(run_input=run_input)
-    dataset_items = await client.dataset(actor_call['defaultDatasetId']).list_items(clean=True)
+    dataset_items = await client.dataset(actor_call['defaultDatasetId']).list_items(clean=True) # type: ignore[index]
     data = dataset_items.items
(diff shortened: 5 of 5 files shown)


pyproject.toml
```diff
@@ -87,3 +87,4 @@ warn_redundant_casts = true
 warn_return_any = true
 warn_unreachable = true
 warn_unused_ignores = true
+ignore_missing_imports = true
```
Pijukatel, line 90: I do not think this is the way to go. We should rather add the packages to dependencies of the project so that type checking works on it, instead of ignoring the types.

MQ37: currently it fails on CrewAI and it does not have any `types-` package. I agree with you, let's ignore the import types for the specific templates instead of globally. What do you think?

MQ37: Some packages like CrewAI do not work with types, will ignore them on template level instead of globally.

Pijukatel:
> For `python-pydanticai` I had to set mypy Python version to `3.10` from `3.9` because of the syntax error in `.venv` which currently cannot be ignored in mypy (at least I could not get that to be ignored). And then disabled imports checks for `python-crewai` and `python-llamaindex-agent` where deps caused issues.

I will start migrating our repos from minimal >= Python3.9 to minimal >=3.10. That will solve some of the issues, but it will take a little bit of time.

Btw. since you also started touching the pydanticai template file, please use the code changes from here instead of ignores:
https://github.com/apify/actor-templates/pull/366/files

MQ37:
> > For `python-pydanticai` I had to set mypy Python version to `3.10` from `3.9` because of the syntax error in `.venv` which currently cannot be ignored in mypy (at least I could not get that to be ignored). And then disabled imports checks for `python-crewai` and `python-llamaindex-agent` where deps caused issues.
> 
> I will start migrating our repos from minimal >= Python3.9 to minimal >=3.10. That will solve some of the issues, but it will take a little bit of time.
> 
> Btw. since you also started touching the pydanticai template file, please use the code changes from here instead of ignores: https://github.com/apify/actor-templates/pull/366/files

Merged the changes for the `agent.py` PydanticAI template, thank you :+1:

.github/workflows/lint_and_test.yaml
```diff
@@ -146,3 +146,53 @@ jobs:
       env:
         APIFY_TOKEN: ${{ secrets.APIFY_TEST_USER_API_TOKEN }}
         OPENAI_API_KEY: ${{ secrets.APIFY_SERVICE_ACCOUNT_OPENAI_API_KEY }}
+
+  test-llm-ai-templates:
+    name: Test LLM AI templates
+    strategy:
+      matrix:
+        os: [ubuntu-22.04, windows-latest]
+        node-version: [ 18, 20, 22 ]
```
Pijukatel, line 155: In this case I think we can just pick one of the node versions.

.github/workflows/lint_and_test.yaml
```diff
@@ -146,3 +146,53 @@ jobs:
...
+        restore-keys: |
+          ${{ runner.os }}-node-${{ matrix.node-version }}
+
+    - name: Install dependencies
+      run: |
+        npm install
+        npm install -g apify-cli
+        ./node_modules/.bin/playwright install chromium --with-deps
+
+    - name: Setup Python ${{ matrix.python-version }}
```
Pijukatel, line 184: python-version is not defined, so it takes whatever default. So I think we should just parametrize over different python versions defined in matrix and with just one specific node version

Pijukatel (commented):
Good, just a comment about test parameters

.github/workflows/lint_and_test.yaml
Pijukatel: Ohh I forgot about the ts templates. So this makes sense, but it should be split for Python llm templates and node llm templates

Pijukatel (commented):
Also it seems that no templates are tested on Windows 
image

and that all llm python templates consistently fail on Linux - which is probably not just flakiness, but something wrong in the test setup

MQ37:
> Also it seems that no templates are tested on Windows image
> 
> and that all llm python templates consistently fail on Linux - which is probably not just flakiness, but something wrong in the test setup

yes, I forgot to split the py and node versions :smile:

.github/workflows/lint_and_test.yaml
MQ37: I personally think there would be too many cicd runs when we would have split that.

.github/workflows/lint_and_test.yaml
MQ37: will add the python version :+1:

MQ37: also will remove node.js 18, since the support ends in 4 weeks

MQ37: agree regarding the versions, let's use single node version and python 3.10+

MQ37:
before merging we just need to change the branch here:

https://github.com/apify/actor-templates/blob/70b9748e67ddddd1cc322d5d99c58f2ba5acb259/.github/workflows/lint_and_test.yaml#L59 

https://github.com/apify/actor-templates/blob/70b9748e67ddddd1cc322d5d99c58f2ba5acb259/.github/workflows/lint_and_test.yaml#L63

https://github.com/apify/actor-templates/blob/70b9748e67ddddd1cc322d5d99c58f2ba5acb259/.github/workflows/lint_and_test.yaml#L67

MQ37:
closes https://github.com/apify/actor-templates/issues/368

Pijukatel (approved):
Thanks. This it is now nicely separated.

B4nan (approved):
only had a brief look, as long as the CI is green, i am fine, but be sure to test the templates once merged

PR #373, opened 2025-04-01
CI test

SuchATestAccount:


B4nan:
i think you need the explicit approval only if its your first PR to given repository, at least that is how it usually works (regardless of the PR target thing)

PR #385, opened 2025-04-02
feat: ts-standby: Update ESLint, Prettier move Dockerfile

oklinov:


templates/ts-standby/.prettierrc
```diff
@@ -0,0 +1 @@
+{ "printWidth": 100 }
```
jirimoravcik, line 1: Hey, why is 100 here, while 120 in `.editorconfig`? Thanks

Also I was told prettier will be only able to either work with `.prettierrc` or `.editorconfig`, not with both

templates/ts-standby/eslint.config.mjs
jirimoravcik, line 1: Did you delete the old eslint config file?

templates/ts-standby/package.json
```diff
@@ -7,23 +7,29 @@
...
         "start:prod": "node dist/main.js",
         "start:dev": "tsx src/main.ts",
         "build": "tsc",
-        "test": "echo \"Error: oops, the actor has no tests yet, sad!\" && exit 1"
+        "test": "echo \"Error: oops, the actor has no tests yet, sad!\" && exit 1",
+        "lint": "eslint .",
+        "lint:fix": "eslint . --fix",
+        "format": "prettier --write .",
+        "format:check": "prettier --check .",
+        "lint:format": "npm run format && npm run lint:fix"
```
jirimoravcik, line 32: I don't really like `lint:format` as it doesn't indicate it's "lint and format". I also wonder which one should be first. Can lint break formatting or formatting break lint? I'd say the first case is more likely, so I'd change the order to 1. lint 2. format.

templates/ts-standby/eslint.config.mjs
```diff
@@ -0,0 +1,33 @@
...
+
+        rules: {
+            // Reason why this is enabled belongs here
+            'max-len': [
+                'error',
+                {
+                    code: 120,
+                    // ignoreComments: true,
+                },
+            ],
```
jirimoravcik, line 30: Honestly, I'd disable this completely as we have the formatter.

jirimoravcik (changes requested):
Thanks, looks great, had a few comments.

templates/ts-standby/eslint.config.mjs
MatousMarik: It is not done in this PR, but the script supports it. I will investigate the reason.

templates/ts-standby/package.json
MatousMarik: I saw it as they both can break each other and since eslint should be correctly configured right now, I wanted to run it last. That is why I chose the name where the first part is "main command" and the second is "option" (in my perception...)

Right now we don't have formatter set up well enough so the code structure is changing a lot from what we are used to.

PR #448, opened 2025-06-25
fix: change ts-mcp-template server to @modelcontextprotocol/server-everything

MQ37:
The @modelcontextprotocol/server-sequential-thinking was confusing for human users, changed MCP server to @modelcontextprotocol/server-everything which is more human user friendly

PR #449, opened 2025-06-26
fix: puppeteer templates that failed on README.md npm run format and ignore venv/.venv and node_modules in eslint config

MQ37:
in my previous PR https://github.com/apify/actor-templates/pull/448 the puppeteer templates were failing on npm run format README.md check I fixed that formatting issue. Also noticed that when running eslint in the root it fails on venv and .venv dirs so I added them to ignore (also added node_modules just in case).

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -5,7 +5,7 @@ import apifyJs from '@apify/eslint-config/js.js';
 
 // eslint-disable-next-line import/no-default-export
 export default [
-    { ignores: ['**/dist', 'templates', 'src'] },
+    { ignores: ['**/dist', '**/venv', '**/.venv', 'templates', 'src', 'node_modules'] },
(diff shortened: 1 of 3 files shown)


eslint.config.mjs
```diff
@@ -5,7 +5,7 @@ import apifyJs from '@apify/eslint-config/js.js';
 
 // eslint-disable-next-line import/no-default-export
 export default [
-    { ignores: ['**/dist', 'templates', 'src'] },
+    { ignores: ['**/dist', '**/venv', '**/.venv', 'templates', 'src', 'node_modules'] },
```
B4nan, line 8: afaik `node_modules` should be ignored by default by eslint

(but lets keep it here, no harm in having it listed explicitly)

B4nan (approved):
thanks!

PR #481, opened 2025-07-17
fix: python-mcp-server comment out env param so that the template works as it is

jirispilka:


jirispilka:
oh, I missed the `or '' `.
Let me close this PR then

PR #490, opened 2025-07-24
feat: Add python crawlee playwright with camoufox template

Pijukatel:
### Description

- Add crawlee + playwright + camoufox template

### Issues

- Closes: #482

### DIFF
--- a/templates/python-crawlee-playwright-camoufox/requirements.txt
+++ b/templates/python-crawlee-playwright-camoufox/requirements.txt
@@ -1,0 +1,6 @@
+# Feel free to add your Python dependencies below. For formatting guidelines, see:
+# https://pip.pypa.io/en/latest/reference/requirements-file-format/
+
+apify < 3.0
+crawlee[playwright]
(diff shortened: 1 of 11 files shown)


templates/python-crawlee-playwright-camoufox/requirements.txt
```diff
@@ -0,0 +1,6 @@
+# Feel free to add your Python dependencies below. For formatting guidelines, see:
+# https://pip.pypa.io/en/latest/reference/requirements-file-format/
+
+apify < 3.0
+crawlee[playwright]
+camoufox[geoip]~=0.4.5
```
vdusek, line 6: Maybe?

```suggestion
camoufox[geoip] < 1.0
```

Pijukatel: Isn't that too risky regarding the 0.x.x initial development versions of Camoufox? Until at least v1 is out, I think it would be safer to allow a smaller set of releases.

Issue #486, opened 2025-07-23
Add OAuth support to MCP templates


MQ37:
Add these router to support OAuth, optinally the favicon.

```
    app.get('/favicon.ico', (_req: Request, res: Response) => {
        res.writeHead(301, { Location: APIFY_FAVICON_URL });
        res.end();
    });

    app.get('/.well-known/oauth-authorization-server', async (_req: Request, res: Response) => {
        // Some MCP clients do not follow redirects, so we need to fetch the data and return it directly.
        const response = await fetch(`${APIFY_API_BASE_URL}/.well-known/oauth-authorization-server`);
        const data = await response.json();
        res.status(200).json(data);
    });
```

PR #494, opened 2025-07-28
feat: enable oauth for mcp templates

MQ37:
closes https://github.com/apify/actor-templates/issues/486

This PR adds endpoints to the MCP server templates to allow user to log in using oauth.


jirispilka (approved):
Thanks!

Issue #487, opened 2025-07-23
Add ability to Actorize streamable MCP server in python-mcp-template


jirispilka:

jirispilka:
> simplest, but not efficent, solution would be to install `npx` and use `mcp-remote`like the TS template does

Good idea.

There will be some overhead in installing npx and mcp-remote, but it would be great to keep both templates as similar as possible. In my opinion, the Python template is more complicated than it needs to be.

jirispilka:
Yes, this was closed in: https://github.com/apify/actor-templates/pull/501

PR #498, opened 2025-07-31
feat(python-mcp-server): remote (streamable) mcp through mcp-remote

MQ37:
Install node runtime with npx and `mcp-remote` package that is used to wrap remote MCP servers as local stdio.

this sadly bumped the image size from ~300 MB to ~600 MB.

closes https://github.com/apify/actor-templates/issues/487


jirispilka:
@MQ37 I believe this could be done in `server.py` using 

from mcp.client.streamable_http import streamablehttp_client
```
        elif self.server_type == ServerType.HTTP:
            params = self.config.model_dump(exclude_unset=True)
            await self._initialize_and_run_server(streamablehttp_client, **params)
```

But I haven't tested that, should I look into that?

jirispilka:
I'm closing this PR in favor of this one: https://github.com/apify/actor-templates/pull/501

PR #507, opened 2025-08-12
feat(python-crewai): update and use crewai integration

MQ37:
Use Apify CrewAI integration instead of creating the tools from scratch.


templates/python-crewai/Dockerfile
```diff
@@ -1,7 +1,11 @@
 # First, specify the base Docker image.
 # You can see the Docker images from Apify at https://hub.docker.com/r/apify/.
 # You can also use any other image from Docker Hub.
-FROM apify/actor-python:3.13
+# We are using Python 3.12 because 3.13 required compilation of the packages for the CrewAI setup.
```
B4nan: @MQ37 would be great if you could resolve this and bump it, dependency updates are blocked by this currently. also can you please check #617 and merge it if it works fine? CI is green, but i'd rather hear it from a human that its ok :]

B4nan: i mean both, the python version is what breaks in the renovate PR, and we also have #617 which bumps crawai itself to v1 

i can imagine it will just work with 3.14, but that's a separate story, lets first unblock update to 3.13, and renovate will let us know if 3.14 works or not

MQ37: the https://github.com/apify/actor-templates/pull/617 works fine with the python 3.12 - at least on the Apify platform it runs

MQ37: there is some issue with CrewAI where it prompts "Would you like to view your execution traces? [y/N]" and prevents the Actor from exiting :facepalm:

PR #514, opened 2025-09-03
fix: Python MCP server stdio, charging, naming

jirispilka:
First, I only wanted to fix stdio, but while I was at it, I also adjusted charging and naming.


jirispilka:
FYI @ivanvia7

PR #525, opened 2025-09-09
chore: Update Python packages constraints

vdusek:


templates/python-crewai/requirements.txt
```diff
@@ -1,4 +1,7 @@
-apify<3.0.0
-apify_client<2.0.0
-langchain-apify<1.0.0
-crewai[tools]<1.0.0
+# Feel free to add your Python dependencies below. For formatting guidelines, see:
+# https://pip.pypa.io/en/latest/reference/requirements-file-format/
+
+apify < 3.0.0
+apify-client
```
B4nan, line 5: this one should be without a version constraint?

vdusek: Yes, because the version of `apify-client` is already managed by the SDK, there's no need to maintain compatibility here as well; that is SDK's responsibility.

PR #532, opened 2025-09-18
feat: Add html comments to the readme

khali1:
In order to detect the readme wasn't changed by the user, we added a comment which should be always present. In the future we will be scanning for such comment in order to determine originality of the readme.

Also added test to ensure new templates will have readmes as well


vdusek (changes requested):
Tests are failing :)

jirispilka (approved):
Looks good (regarding the changes in the MCP template. Thanks!)

Jkuzz (approved):
lgtm

PR #538, opened 2025-09-25
fix: Remove double charing in ts-mcp template

jirispilka:
FYI: @MQ37


jirispilka:
> I am wondering if there was a reason for these charging calls at the first place
We should update the template to remove it 


>  or if it is a product of vibe-coding 🤔
@MQ37 can tell :)

jirispilka:
`ts-crawlee-playwright-camoufox` is failing, the test is flaky, merging anyway 🤞🏻

PR #572, opened 2025-10-24
feat: Add output schema for all templates

patrikbraborec:


B4nan (approved):
needs reformatting, i guess because of the double blank line in many places. approving anyway, as long as the CI will be green

valekjo (approved):
Looks good. Didn't go through all of the files though, seems the changes are very similar everywhere.

PR #582, opened 2025-11-11
fix: python-crawlee-playwright-camoufox command is using wrong template

ivanvs:
In the current version when you try to use python-crawlee-playwright-camoufox you will get python-crawlee-playwright template. This PR is fixing this problem.


B4nan (approved):
good catch, thanks!

PR #584, opened 2025-11-18
chore(ci): skip `ts-playwright-test-runner` CI tests

barjin:
`ts-playwright-test-runner` template test fails on `npm install` under Windows 11 / Node 24 with `ECOMPROMISED` ([see e.g. here](https://github.com/apify/actor-templates/actions/runs/19433613745/job/55598569066#step:6:1352)).

This PR skips this test in CI for the time being.

Possibly related to https://github.com/actions/runner-images/issues/13238

### DIFF
--- a/templates/manifest.json
+++ b/templates/manifest.json
@@ -499,6 +499,7 @@
             "label": "Playwright + Chrome Test Runner",
             "category": "typescript",
             "technologies": ["nodejs", "playwright", "chrome"],
+            "skipTests": true,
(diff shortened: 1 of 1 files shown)


B4nan:
Hmm, actually, the tests stopped failing, apparently?

barjin:
I don't know about that, see https://github.com/apify/actor-templates/actions/runs/19433613745/job/55598569066#step:6:1352 16 hours ago.

It's curious that the past 2 scheduled ones passed, but I'd chalk that up to randomness rather than an actual fix getting implemented... wdyt?

PR #607, opened 2025-12-31
ci: update package.json when triggering PR

vladfrangu:


B4nan:
can you also try updating npm to resolve that ECOMPROMISED windows issue?

PR #690, opened 2026-02-18
feat: Add CLAUDE.md symlinks to AGENTS.md in all templates

patrikbraborec:
## Summary
- Updated `scripts/copy-agents-md-to-templates.mts` to create a `CLAUDE.md` symlink pointing to `AGENTS.md` in each template directory
- Each template now ships with both `AGENTS.md` (source of truth) and `CLAUDE.md` (symlink) for compatibility with Claude Code
- Symlink creation is idempotent — existing `CLAUDE.md` files are removed before recreating

### DIFF
--- a/scripts/copy-agents-md-to-templates.mts
+++ b/scripts/copy-agents-md-to-templates.mts
@@ -1,4 +1,4 @@
-import { readdir, readFile, writeFile } from 'node:fs/promises';
+import { readdir, readFile, writeFile, symlink, unlink, lstat } from 'node:fs/promises';
 
 const baseDir = new URL('../agent-bases/', import.meta.url);
 const templatesDir = new URL('../templates/', import.meta.url);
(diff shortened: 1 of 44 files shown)


vladfrangu (commented):
Will this work with CLI? we can use the hidden --template-archive-url flag

patrikbraborec:
@vladfrangu It will work on CI. The build script `src/build_templates.js` runs zip -X without the -y flag, so symlinks are dereferenced -> CLAUDE.md ends up as a full copy of AGENTS.md in the zip, not a symlink.

Users running `apify create --template-archive-url` will get a working CLAUDE.md with the correct content.

Only downside: archives get slightly larger (~22KB duplicate per template), and the two files won't be linked -> edits to one won't reflect in the other, which is IMHO ok as it will be users decision.

vladfrangu (approved):

PR #695, opened 2026-02-20
ci: create PR and automatically merge when a dependency updates

vladfrangu:
This change will keep opening the PR for new image versions but will automatically merge it


vladfrangu:
I do like the idea of PRs with auto-merge tho, might to that route

PR #698, opened 2026-02-27
chore: Remove view param from output_schema.json

patrikbraborec:
## Summary
- Removed `?view=overview` query parameter from the `template` URL in all `output_schema.json` files across 29 templates
- Renamed the output property from `overview`/`Overview` to `results`/`Results` to match the simplified default
- The simple `/items` endpoint is sufficient for most Actors that store data in the default dataset

### DIFF
--- a/templates/ts-playwright-test-runner/.actor/output_schema.json
+++ b/templates/ts-playwright-test-runner/.actor/output_schema.json
@@ -2,15 +2,10 @@
     "actorOutputSchemaVersion": 1,
     "title": "Output schema",
     "properties": {
-        "overview": {
+        "results": {
             "type": "string",
-            "title": "Overview",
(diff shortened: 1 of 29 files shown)


gippy (approved):
Thanks 🙂

templates/ts-playwright-test-runner/.actor/output_schema.json
```diff
@@ -2,15 +2,10 @@
...
-        "overview": {
+        "results": {
             "type": "string",
-            "title": "Overview",
-            "template": "{{links.apiDefaultDatasetUrl}}/items?view=overview"
-        },
-        "report": {
-            "type": "string",
-            "title": "HTML Report",
-            "template": "{{links.apiDefaultKeyValueStoreUrl}}/records/report"
```
gippy: I think here you could have kept the link to the key value store.

PR #738, opened 2026-03-20
chore: drop Python 3.10 support

vdusek:
## Summary

- Remove Python 3.10 from CI test matrices
- Bump `requires-python` to `>= 3.11`
- Uncomment `crewai` and `arxiv-mcp-server` deps (required Python 3.11+)
- Use `StrEnum` instead of `(str, Enum)` in mcp-proxy template

## Test plan

- [x] CI passes on Python 3.11-3.14


pyproject.toml
```diff
@@ -1,21 +1,23 @@
...
+    # "crewai[tools] >= 1.12.2, < 2.0.0",
     "fastapi >= 0.135.0, < 1.0.0",
-    "fastmcp>=3.2.0,<4.0.0",
+    "fastmcp >= 3.0.0, < 4.0.0",
     "httpx >= 0.28.0, < 1.0.0",
     "langchain >= 1.0.0, < 2.0.0",
-    # "langchain-apify >= 0.1.0, < 1.0.0",  # Incompatible with langchain >= 1.0.0 (requires langchain-core < 0.4.0)
-    "langchain-openai>=1.1.14,<2.0.0",
+    # langchain-apify conflicts with langchain >= 1.0.0 (requires langchain-core < 0.4.0)
+    # "langchain-apify >= 0.1.0, < 1.0.0",
```
Pijukatel, line 19: What is the point of those commented-out packages?

Pijukatel (approved):

pyproject.toml
vdusek: It is so that we know these can't currently be installed in the same environment and therefore can't be linted or type-checked. Or do you mean something else?

vdusek:
The failing checks are pre-existing infra issue - OpenAI service-account key is being rejected.

PR #760, opened 2026-04-28
fix(ts-templates): disable incremental builds

### LINKED ISSUE #754: Crawlee+Cheerio template missing dist folder
I found a bug involving the latest Crawlee+Cheerio template, which uses TypeScript 6 (I suspect the migration caused this bug).
The problem is that, after building the actor, a tsconfig.tsbuildinfo file is generated in the root of the project.
If I then delete the dist folder, the next build won't generate anything.
The way I noticed is that I tried building the actor locally, pushed it to the platform with the CLI, and failed to run it, because the dist folder was missing.

A possible workaround would be to add "tsBuildInfoFile": "dist/tsconfig.tsbuildinfo" to tsconfig.json.

Steps to reproduce:
```
apify create test --template=ts-crawlee-cheerio
cd test
npm install
npm run build
rm -rf dist
npm run build
```
Now no dist folder will be generated.

Alternatively:
```
apify create test --template=ts-crawlee-cheerio
cd test
npm install
npm run build
apify push
```
The next run on the platform will fail.

Reported on [slack](https://apify.slack.com/archives/C0L33UM7Z/p1775581620813169)

B4nan:
## Summary

- All `templates/ts-*/tsconfig.json` files now set `"incremental": false` so `tsc` no longer writes a `tsconfig.tsbuildinfo` file at all.
- Adds `test/tsconfig.test.js` as a config-shape regression guard for all 14 TypeScript templates.

## Why

`@apify/tsconfig` sets `incremental: true`. Without an explicit `tsBuildInfoFile`, `tsc` writes `tsconfig.tsbuildinfo` next to `tsconfig.json` (the project root). After a user runs `rm -rf dist`, the buildinfo at the root survives and convinces `tsc` that nothing changed, so the next `npm run build` emits nothing — breaking `apify push` (uploads with no `dist/`) and `node dist/main.js`.

The first version of this PR fixed it by moving the buildinfo into `outDir`, but that ships the buildinfo file inside the actor. Templates are built once on `apify push`, so the speedup from incremental compilation is negligible — disabling it removes the buildinfo file entirely and eliminates the class of bug rather than papering over one symptom.

Reproduction (from #754):
```
apify create test --template=ts-crawlee-cheerio
cd test && npm install && npm run build
rm -rf dist && npm run build  # before fix: no dist/ generated
```

Closes #754

## Test plan

- [x] Reproduce the bug on `ts-crawlee-cheerio` before the fix (no `dist/` after second build)
- [x] Verify the fix on `ts-crawlee-cheerio`: `npm run build`, `rm -rf dist`, `npm run build` → `dist/` is regenerated and no `tsconfig.tsbuildinfo` is created (in `dist/` or at the repo root)
- [x] `npm run test-without-templates` passes (17 tests, includes the new 14)
- [x] `npm run lint` passes
- [ ] CI runs the heavy `test-templates` job against all TS templates

### DIFF
--- a/templates/ts-beeai-agent/tsconfig.json
+++ b/templates/ts-beeai-agent/tsconfig.json
@@ -6,6 +6,7 @@
         "target": "ES2022",
         "outDir": "dist",
         "rootDir": "./src",
+        "tsBuildInfoFile": "dist/tsconfig.tsbuildinfo",
         "noUnusedLocals": false,
         "skipLibCheck": true,
         "lib": ["DOM"]
--- a/templates/ts-beeai-agent/tsconfig.json
+++ b/templates/ts-beeai-agent/tsconfig.json
@@ -6,7 +6,7 @@
         "target": "ES2022",
         "outDir": "dist",
         "rootDir": "./src",
-        "tsBuildInfoFile": "dist/tsconfig.tsbuildinfo",
+        "incremental": false,
         "noUnusedLocals": false,
         "skipLibCheck": true,
         "lib": ["DOM"]
--- a/templates/ts-bootstrap-cheerio-crawler/tsconfig.json
+++ b/templates/ts-bootstrap-cheerio-crawler/tsconfig.json
@@ -6,6 +6,7 @@
         "target": "ES2022",
         "outDir": "dist",
         "rootDir": "./src",
+        "tsBuildInfoFile": "dist/tsconfig.tsbuildinfo",
         "noUnusedLocals": false,
         "skipLibCheck": true,
         "lib": ["DOM"]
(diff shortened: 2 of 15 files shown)


barjin (approved):
lgtm, thank you @B4nan !

----

One small nit is that the `tsbuildinfo` file will end up in the built Actor now. I suppose this doesn't really matter (unless you're doing some hardcore optimisations, in which case you wouldn't use the default templates anyway). Good to go by me 👍

B4nan:
Hmm, true that, I'll disable the incremental flag in the templates instead, it doesn't really make much sense here.

PR #770, opened 2026-05-14
chore: Use custom action to commit changes in CI instead of `git commit`

fnesveda:
We want to enforce commit signing for all commits in our repositories.
To do that, we need to make sure even commits created by CI workflows are signed.

It would be possible to sign using GPG keys, but that would require a lot of maintenance.

Instead, we can commit using the GitHub GraphQL API, which automatically signs commits.

This PR replaces direct `git commit` / `git push` usage (and third-party commit actions like `EndBug/add-and-commit`)
with the `apify/actions/signed-commit` action, which uses the GraphQL API under the hood.


B4nan (approved):

vladfrangu (commented):
For the update-templates action, can you manually run it for, say, node templates (or I can on monday) to check if it works now 🙏

fnesveda:
> For the update-templates action, can you manually run it for, say, node templates (or I can on monday) to check if it works now 🙏

It works: https://github.com/apify/actor-templates/actions/runs/25969171705/job/76337933333

But I almost updated the templates in production, the generated PR has automerge turned on, good thing I noticed 😄

vladfrangu (approved):

PR #803, opened 2026-06-04
chore: Update templates for base image: apify/actor-node

apify-service-account:
Automated update of templates for base image `apify/actor-node`

**Parameters:**
- Base Image: `apify/actor-node`
- Runtime Version: `24`
- Module Version: `N/A`

> Generated by [Update Templates](https://github.com/apify/actor-templates/actions/workflows/update-templates.yml) workflow.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,3 +1,202 @@
+---
+lockfileVersion: '9.0'
+
+importers:
+
+  .:
+    configDependencies: {}
+    packageManagerDependencies:
+      '@pnpm/exe':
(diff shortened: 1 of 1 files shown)


B4nan:
@vladfrangu the lockfile diff looks completely off in here, any idea why?

vladfrangu:
It used pnpm 11.. We're still on 10 right?

B4nan:
nope https://github.com/apify/actor-templates/blob/master/package.json#L58

also the lockfile versions match with what was there before, it just prepended stuff...

PR #824, opened 2026-06-27
chore: audit and refine template technology tags

DaveHanns:
## What & why

Addresses the **tags** part of #763 (_"Audit `tags` — remove stale ones, add missing ones, ensure consistency"_). This is the `category` (language) and `technologies` fields only — it deliberately does **not** touch `label`/`description`, which are handled by #814. The two PRs are independent (both edit `templates/manifest.json` but different fields, so expect trivial merge conflicts whichever lands second). `useCases` were already redacted separately and are untouched here.

> **Companion PR:** apify/apify-core#28932 adds the missing Console icons (`logo_camoufox.svg`, `logo_langchain.svg`, `logo_parsel.svg`) for the tags kept below.

## How the audit was done

Every template's `technologies` was checked against (a) its **real dependencies/source** on disk and (b) the Console's **icon set**. Each tag (and the `category`) renders in the Console as `/img/logos_other/logo_<value>.svg` with **no `onError` fallback** — so any tag without a matching icon shows a *broken image* in the template gallery, the card, and the sign-up header.

The audit confirmed that the **majority of tags are already accurate**. The changes below are the genuine inconsistencies/inaccuracies found.

## Changes (9 templates, `technologies` only)

| Template | Before | After | Reason |
|---|---|---|---|
| `ts-beeai-agent` | `[beeai]` | `[nodejs, beeai]` | Runtime tag missing — every other JS/TS template lists `nodejs` |
| `ts-mastraai` | `[mastraai]` | `[nodejs, mastraai]` | ″ |
| `ts-mcp-proxy` | `[mcp]` | `[nodejs, mcp]` | ″ |
| `ts-mcp-empty` | `[mcp]` | `[nodejs, mcp]` | ″ |
| `ts-playwright-test-runner` | `[nodejs, playwright, chrome]` | `[nodejs, playwright]` | **Inaccurate** — runs the all-browsers Playwright image (Chromium/Firefox/WebKit), not Chrome-only (`Dockerfile`: `apify/actor-node-playwright:24-…`, README §"No more problems with incompatible browser versions") |
| `js-start` | `[nodejs, cheerio, axios]` | `[nodejs, cheerio]` | `axios` has no icon; `cheerio` is the headline tech |
| `ts-start` | `[nodejs, cheerio, axios]` | `[nodejs, cheerio]` | ″ |
| `ts-start-bun` | `[bun, cheerio, axios]` | `[bun, cheerio]` | ″ |
[...]

PR #827, opened 2026-06-29
build: move repo to prettier 3.9.x (fix template format:check drift)

DaveHanns:
## Root cause: prettier version drift between root and templates

The per-template `format:check` (run in CI by **Test Node.js templates** / **Test LLM AI Node.js templates**) started failing on every PR run on/after **2026-06-27** (e.g. [#824](https://github.com/apify/actor-templates/pull/824)) due to a **version mismatch**:

| | prettier range | lockfile | version it installs |
|---|---|---|---|
| **repo root** | `^3.7.4` | ✅ `pnpm-lock.yaml` | **3.8.3** (frozen) |
| **each template** | `^3.5.3` | ❌ none | **3.9.x** (fresh `npm install` floats to newest) |

**Prettier 3.9.0** (2026-06-27) reformats the comment-only objects in the `AGENTS.md` input/output schema snippets. So the templates' 3.9.x flagged files that the root's pinned 3.8.3 considers correctly formatted.

## Fix: put the whole repo on prettier 3.9.x
Bump the root prettier and reformat, so the root and the templates' floating installs agree.

## Notes
- Root lands on **3.9.1**, not 3.9.3 — the repo's `minimumReleaseAge: 1440` (24h) blocks 3.9.2/3.9.3 (released today). The templates' npm install has no such policy and floats to 3.9.3; 3.9.1 and 3.9.3 produce identical output for these files (verified). They'll converge once 3.9.3 is >24h old.

- The templates still float (`^`, no lockfile), so a *future* prettier release that changes formatting can re-introduce this for the templates (the root is shielded by its lockfile until bumped deliberately).

### DIFF
--- a/agent-bases/js.AGENTS.md
+++ b/agent-bases/js.AGENTS.md
@@ -220,9 +220,7 @@ The input schema defines the input parameters for an Actor. It's a JSON object c
     "title": "<INPUT-SCHEMA-TITLE>",
     "type": "object",
     "schemaVersion": 1,
-    "properties": {
-        /* define input fields here */
-    },
+    "properties": {/* define input fields here */},
     "required": []
 }
 ```
@@ -286,9 +284,7 @@ The Actor output schema builds upon the schemas for the dataset and key-value st
 {
     "actorOutputSchemaVersion": 1,
     "title": "<OUTPUT-SCHEMA-TITLE>",
-    "properties": {
-        /* define your outputs here */
-    }
+    "properties": {/* define your outputs here */}
 }
 ```
 
--- a/agent-bases/python.AGENTS.md
+++ b/agent-bases/python.AGENTS.md
@@ -222,9 +222,7 @@ The input schema defines the input parameters for an Actor. It's a JSON object c
     "title": "<INPUT-SCHEMA-TITLE>",
     "type": "object",
     "schemaVersion": 1,
-    "properties": {
-        /* define input fields here */
-    },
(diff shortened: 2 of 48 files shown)


agent-bases/js.AGENTS.md
```diff
@@ -220,9 +220,7 @@ The input schema defines the input parameters for an Actor. It's a JSON object c
     "title": "<INPUT-SCHEMA-TITLE>",
     "type": "object",
     "schemaVersion": 1,
-    "properties": {
-        /* define input fields here */
```
vladfrangu, line 224: We probably don't want this collapsed right?

vladfrangu (approved):

agent-bases/js.AGENTS.md
DaveHanns: I think it does not really matter much for the agents.

Just for a record, this was the issue failing the tests. The older version of prettier (root level) kept it spread, but the newer versions of prettier (template level) marked it as incorrect formatting.
Now both root and template level are at versions that collapse it.

fisker: You can use `//` comment if you want force it expanded

DaveHanns: Thanks @fisker, I was considering it as a possible solution, but decided to not clutter the md files if not really necessary. As these are supposed to be read mainly by agents, the one-liner is not really an issue.

l2ysho (approved):

PR #833, opened 2026-07-04
fix(ts-templates): move tsx and typescript to dependencies

DaveHanns:
## Summary

Every ts-* template declares `tsx` and `typescript` in `devDependencies`, but the template's own runtime scripts require both:

```json
"start:dev": "tsx src/main.ts",
"build": "tsc"
```

If `npm install` runs in an environment where devDependencies are skipped (`NODE_ENV=production`, `npm install --omit=dev`, npm 11's tighter production defaults, or any parent installer that forwards those flags), a freshly scaffolded project fails immediately.

This PR moves `tsx` and `typescript` from `devDependencies` to `dependencies` in the 12 ts-* templates whose `start:dev`/`build` scripts invoke them. `ts-start-bun` (runs `bun src/main.ts` directly) and `ts-playwright-test-runner` (already lists both in `dependencies`) are left unchanged.

## Reproducer

```console
$ NODE_ENV=production apify create tmp -t ts-empty
$ cd tmp && apify run
sh: tsx: not found
Error: /usr/local/bin/npm exited with code 127

$ npm run build
sh: tsc: not found
# `npx tsc` then quietly installs the deprecated `tsc@2.0.4` squatter
# ("This is not the tsc command you are looking for")
```

The `.bin/tsx` and `.bin/tsc` shims are absent because npm never installed them:

```console
$ ls node_modules/.bin/tsx
ls: cannot access 'node_modules/.bin/tsx': No such file or directory
$ ls node_modules/typescript
ls: cannot access 'node_modules/typescript': No such file or directory
```

Workarounds users currently need to discover on their own:

```console
$ NODE_ENV=development npm install
# or
$ npm install --include=dev
```

## Why this is safe for image size

The Dockerfile builder stage already passes `--include=dev`, so image builds are unaffected. The runtime image copies the compiled `dist/` from the builder and installs with `--omit=dev --omit=optional` — neither `tsx` nor `typescript` is ever installed into the runtime image today, and this PR does not change that (both remain absent because the runtime image starts from `dist/main.js`, which needs neither). Only the local `apify run`/`npm run build` flows benefit.

## Changes

- `templates/ts-empty/package.json` — move `tsx`, `typescript` to `dependencies`
- `templates/ts-crawlee-cheerio/package.json` — same
- `templates/ts-crawlee-playwright-chrome/package.json` — same
- `templates/ts-crawlee-playwright-camoufox/package.json` — same
- `templates/ts-crawlee-puppeteer-chrome/package.json` — same
- `templates/ts-standby/package.json` — same
- `templates/ts-start/package.json` — same
- `templates/ts-bootstrap-cheerio-crawler/package.json` — same
- `templates/ts-mcp-empty/package.json` — same
- `templates/ts-mcp-proxy/package.json` — same
- `templates/ts-beeai-agent/package.json` — same
- `templates/ts-mastraai/package.json` — same

Unchanged:

- `templates/ts-start-bun/package.json` — `start` uses `bun src/main.ts` directly; `tsx`/`typescript` genuinely are devDeps here.
- `templates/ts-playwright-test-runner/package.json` — already lists both under `dependencies`.

## Test plan

- [ ] `NODE_ENV=production apify create tmp -t ts-empty && cd tmp && apify run` exits 0.
- [ ] `NODE_ENV=production npm run build` succeeds and emits `dist/` for every changed template.
- [ ] Existing Jest template suite still passes (`pnpm run test-node-templates`).
- [ ] Docker image build (unchanged Dockerfile) still succeeds and produces the same runtime image contents.

_Surfaced during an evaluation of Apify surfaces for agent-driven Actor development._

### DIFF
--- a/templates/ts-beeai-agent/package.json
+++ b/templates/ts-beeai-agent/package.json
@@ -10,7 +10,9 @@
         "@ai-sdk/openai": "3.0.63",
         "apify": "^3.7.2",
         "beeai-framework": "^0.1.29",
-        "zod": "^3.25.76"
+        "zod": "^3.25.76",
+        "tsx": "^4.22.3",
+        "typescript": "^6.0.3"
     },
     "devDependencies": {
         "@apify/eslint-config": "^2.0.6",
@@ -20,8 +22,6 @@
         "eslint-config-prettier": "^10.1.8",
         "globals": "^17.6.0",
         "prettier": "^3.8.3",
-        "tsx": "^4.22.3",
-        "typescript": "^6.0.3",
         "typescript-eslint": "^8.60.0",
         "vitest": "^4.1.7"
     },
--- a/templates/ts-bootstrap-cheerio-crawler/package.json
+++ b/templates/ts-bootstrap-cheerio-crawler/package.json
@@ -8,7 +8,9 @@
     },
     "dependencies": {
         "apify": "^3.7.0",
-        "@crawlee/cheerio": "^3.15.3"
+        "@crawlee/cheerio": "^3.15.3",
+        "tsx": "^4.20.3",
+        "typescript": "^6.0.0"
     },
     "devDependencies": {
         "@apify/eslint-config": "^2.0.0",
@@ -18,8 +20,6 @@
         "eslint-config-prettier": "^10.1.5",
         "globals": "^17.0.0",
         "prettier": "^3.5.3",
-        "tsx": "^4.20.3",
-        "typescript": "^6.0.0",
         "typescript-eslint": "^8.58.0",
         "vitest": "^4.1.0"
     },
--- a/templates/ts-crawlee-cheerio/package.json
+++ b/templates/ts-crawlee-cheerio/package.json
@@ -8,7 +8,9 @@
     },
     "dependencies": {
         "apify": "^3.7.0",
-        "@crawlee/cheerio": "^3.15.3"
+        "@crawlee/cheerio": "^3.15.3",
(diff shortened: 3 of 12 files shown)


DaveHanns:
Closing — moving `tsx`/`typescript` to `dependencies` fixes the local-dev failure but bloats every ts-* Actor's runtime image by ~34MB (empirically measured: 47MB → 81MB `node_modules` after `npm install --omit=dev --omit=optional` in the runtime stage). The compiled `dist/main.js` never needs these packages at runtime, so the added size is 100% waste.

The "safe for image size" claim in the original PR body is incorrect — the runtime stage's `--omit=dev` strips devDependencies but installs everything under `dependencies`, so moving them across the boundary does land them in the final image.

Better fixes for the same symptom, no image bloat:

- [apify/apify-cli#1261](https://github.com/apify/apify-cli/pull/1261) — force `--include=dev` + `NODE_ENV=development` on the post-scaffold install so a freshly-created ts-* project has the runner binaries available for `apify run` / `npm run build`, plus a preflight for the missing binary case.
- [apify/apify-cli#1252](https://github.com/apify/apify-cli/pull/1252) — on `apify push`, detect a Dockerfile that runs `npm install --omit=dev` before `npm run build` (which invokes `tsc`) and warn before the remote build burns a cycle.
- [apify/apify-docs#2716](https://github.com/apify/apify-docs/pull/2716) — docs callout on the Source-code page for the search-fallback / direct-`docker-build` case.

PR #787, opened 2026-05-28
fix(python-pydanticai): bump pydantic-ai for SSRF advisories and migrate LLM to Apify OpenRouter proxy

MQ37:
## Context

`pydantic-ai >= 1.0.0` in `pyproject.toml` and `templates/python-pydanticai/requirements.txt` resolves to versions vulnerable to two SSRF advisories ([GHSA-cqp8-fcvh-x7r3](https://github.com/pydantic/pydantic-ai/security/advisories/GHSA-cqp8-fcvh-x7r3), [GHSA-2jrp-274c-jhv3](https://github.com/pydantic/pydantic-ai/security/advisories/GHSA-2jrp-274c-jhv3)). The template also forces users to provide their own `OPENAI_API_KEY`, even though they're paying with Apify credits.

## Solution

Bump the floor to `>= 1.99.0` (covers both SSRF advisories) and migrate the template's LLM calls to the [Apify OpenRouter proxy](https://apify.com/apify/openrouter) via `OpenAIChatModel` + `OpenAIProvider(base_url=…, api_key=APIFY_TOKEN)` — same pattern as the [BeeAI template](https://github.com/apify/actor-templates/pull/784).

## Worth your attention

- **`api_key=os.environ['APIFY_TOKEN']`** in `agents.py` — the OpenAI client serializes `api_key` as `Authorization: Bearer …`, which is exactly what the proxy expects. No need for a custom `AsyncOpenAI` client with `default_headers`.
- **14-day freshness rule deviated** — `pydantic-ai 1.99.0` is 8 days old. The newest patched version is required to address the advisories; no patched version is older than 14 days. Standard security-CVE override.
- **`Deps` dataclass grew a `model_name` field** so the nested `create_joke` tool agent uses the same model as the outer one.
- **`modelName` enum mirrors the BeeAI template** for cross-template consistency (`deepseek/deepseek-v4-flash`, `deepseek/deepseek-v4-pro`, `moonshotai/kimi-k2.6`, `z-ai/glm-5.1`).
- **Lockfile churn** — `uv lock` updated transitive deps (`openai`, `mistralai`, `google-genai`, `temporalio`) pulled in by the new `pydantic-ai-slim 1.103.0`; restricting `--upgrade-package` to the pydantic-ai family didn't avoid the cascade because those upgrades are part of the resolver's closure.
- **README** — pre-existing wrong link target (`src/agents.py`) fixed to `my_actor/agents.py`.

PR #842, opened 2026-07-07
test: skip local actor run for OpenRouter-proxy LLM templates

l2ysho:
#784 (ts-beeai-agent), #787 (python-pydanticai) broke the [test template job](https://github.com/apify/actor-templates/actions/runs/28833747881)
 
 - all tests are local, not on platform (no logged in user here) so OpenRouter is not available @jirispilka @MQ37 this is why tests were failing in your PRs. I skipped the run because in my opinion it is too big change to try exercise this on platform (flakyness) but I am open to discuss.
 - other tests failing in playwright, still investigating, but it seems playwright fail on `Uncaught Error: Minified React error #418` error when access apify.com


MQ37 (approved):
Makes sense! Thank you, I did not notice that.

templates/js-crawlee-playwright-camoufox/src/main.js
```diff
@@ -7,14 +7,27 @@
...
+// error). Survive that one specific decode bug so a broken page can't kill the whole
+// run; anything else stays fatal. Remove once the upstream playwright bug is fixed.
+process.on('uncaughtException', (err) => {
+    if (err instanceof TypeError && err.message.includes("reading 'url'") && (err.stack ?? '').includes('coreBundle')) {
+        log.warning(`Ignored known playwright Firefox pageError decode bug: ${err.message}`);
+        return;
+    }
+    throw err;
+});
+
```
l2ysho: this is temporary, lets wait for release of  https://github.com/daijro/camoufox/pull/625 (v152.0.2-alpha), or probably better to pin playwright back to 1.59.x in the camoufox templates 🤔

l2ysho: Went with the pin: dropped the `uncaughtException` guard and pinned `playwright` to `1.59.1` (plus base image `apify/actor-node-playwright-camoufox:24-1.59.1`) in both camoufox templates.

Context:
- playwright 1.60+ crashes with `TypeError: Cannot read properties of undefined (reading 'url')` — the Firefox session decode does an unguarded `pageError.location.url` when a page's uncaught error has no source location (e.g. React minified error #418 on apify.com). One broken page kills the whole crawler run.
- 1.59.1 is the last version without that decode path, so the templates work out of the box again. Exact pin (not `1.59.x`) to keep npm and the Docker base image in lockstep.
- Temporary: unpin once https://github.com/daijro/camoufox/pull/625 (v152.0.2-alpha) / the upstream playwright fix is released. Note the automated base-image bump job has rewritten these versions before, so watch it doesn't re-bump past the pin before then.

jirispilka:
oh, thanks a lot @l2ysho !

PR #844, opened 2026-07-08
fix: add missing Chrome technology to python-crawlee-playwright template

vdusek:
The `python-crawlee-playwright` template is labeled "Crawlee + Playwright + Chrome" and its JS/TS siblings list `chrome` among their technologies, but the Python one was missing it. As a result, the Chrome icon didn't render on the template card on https://apify.com/templates.

Adding `chrome` to the `technologies` array so the icon shows, matching the label and the JS/TS templates.

The web frontend reads this manifest live from the `master` branch, so the icon appears once this is merged. Companion icon work (Camoufox, Parsel) lives in an apify-web PR.


l2ysho (approved):
🚀

PR #904, opened 2026-08-31
chore: pin apify to 4.0.2b5 beta in Python templates

vdusek:
Pins `apify` to the `4.0.2b5` beta (exact pin — pip does not select pre-releases from a version range) in all Python templates, to run the template smoke tests against the upcoming SDK release. Covers 17 `requirements.txt` templates plus `python-uv` (`pyproject.toml` + `uv.lock`).

`python-crewai` is left unchanged: it is still on SDK v3 because `langchain-apify` caps `apify-client < 3.0.0`, which is incompatible with apify 4.x (requires `apify-client >= 3.1.0`).

Not meant to be merged as-is — testing only.

*✍️ Drafted by Claude Code*


vdusek:
just testing PR

PR #907, opened 2026-09-02
ci: use the fast disk and skip --with-deps on the Windows legs

l2ysho:
> [!IMPORTANT] 
> **TL;DR**
> Results from long claude CI debugging session (Windows tests):
>  - it seems it is important to pick right disk drive (in this scenario D, more info below)
> - --with-deps flag in win install also **Media Foundation** Windows feature 
> - longest running tests is now ~12m (vs 30m)

## What was broken

CI tests each template by installing it and running it, on Linux and on Windows.

The Windows legs got slow enough that a job hit `timeout-minutes: 30` and was killed. No test failed — the clock ran out mid-install. That blocked #814. A second leg died the same way on the nightly master run [33575265665](https://github.com/apify/actor-templates/actions/runs/33575265665), on a different shard and a different Node version, so it wasn't one bad shard.

Windows legs had no margin left: successes were already landing at 24–29m under a 30m cap.

The same work took ~4m on Linux and ~20m on Windows.

## Cause 1: the tests were writing to the slow disk

These runners have two drives, and they are not the same kind of thing:

- **`C:`** is not really in the machine. It is storage elsewhere in the datacenter, reached over the network.
- **`D:`** is a real SSD attached to the host.

Installing a template writes tens of thousands of tiny files. On `C:` each one waits for an acknowledgement to come back over the network — you pay that delay per file, not per megabyte.

CI already puts most things on `D:` (`RUNNER_TEMP`, the workspace, the pnpm store). But `prepareActor` in `test/templates.test.js` installs into `os.tmpdir()`, and on Windows that resolves to `C:\Users\...\AppData\Local\Temp`. So the single most expensive part of the job ran on the slowest disk.

**Evidence.** A throwaway benchmark ran the same workload against both paths on four Windows legs — 10k small files in a `node_modules`-shaped tree, plus one 64 MB sequential write:

| | files/s | bulk MB/s |
| --- | --- | --- |
| `os.tmpdir()` → `C:\Users\...\Local\Temp` | 744–1140 | 55–202 |
| `RUNNER_TEMP` → `D:\a\_temp` | 4295–6008 | 1012–1365 |
| Linux `/tmp`, for reference | 27539–32663 | ~1500 |

4.2–5.9x on file creates. The Ubuntu legs were the control — both paths share a disk there and came out at 1.13–1.16x, so the benchmark measured what it claimed. Free space on `D:` came back at 145.9 GiB, so there is no headroom problem.

## Cause 2: `--with-deps` means something different on Windows

One step installs a browser: `playwright install chromium --with-deps`. That flag means "also install what the OS needs".

On Linux it installs a few apt libraries Chromium needs. Cheap — the whole step is 0.4m.

On Windows it enables the **Media Foundation** Windows feature, which provides video codecs. That took **3m16s**, every job, every run. The browser downloads themselves take 15s.

These templates crawl web pages. They never play video.

**Evidence** — same leg, `windows-latest, 22, 1/4`, before and after:

```
# before (run 33680444064)
20:41:30  ##[group]Run pnpm exec playwright install chromium --with-deps
20:41:32  Done in 648ms using pnpm v11.13.0
                                              <- 3m16s, nothing logged
20:44:48  Success Restart Needed Exit Code      Feature Result
20:44:48  True    No             Success        {Media Foundation}
20:44:48  Downloading Chrome for Testing 149.0.7827.55 ... 183.6 MiB
20:45:03  step ends

# after (run 33683171180)
21:06:53  ##[group]Run pnpm exec playwright install chromium
21:06:55  Done in 510ms using pnpm v11.13.0
21:06:55  Downloading Chrome for Testing 149.0.7827.55 ... 183.6 MiB
21:07:11  step ends
```

Identical pnpm resolve, identical 183.6 MiB download. The only difference is the block between them. `Media Foundation` appears zero times in the after-log, and the step went 3.34m → 0.36m across all 8 legs with no overlap.

All 16 legs pass without it, so nothing here needed those codecs.

## The change

Two things, workflow YAML only. Nothing in the templates or the test harness.

```yaml
- name: Install templates on the runner's fast disk
  shell: bash
  run: |
    echo "TMP=$RUNNER_TEMP" >> "$GITHUB_ENV"
    echo "TEMP=$RUNNER_TEMP" >> "$GITHUB_ENV"
```

```yaml
- name: Install playwright browsers
  run: pnpm exec playwright install chromium ${{ runner.os == 'Linux' && '--with-deps' || '' }}
```

The first has to come from a step, not job-level `env`, because the `runner` context is not available there. Both are no-ops on Linux.

## Results

Windows legs of `Test Node.js templates`, all 16 jobs green on run [33683171180](https://github.com/apify/actor-templates/actions/runs/33683171180):

| | playwright step | harness | whole job | worst job |
| --- | --- | --- | --- | --- |
| baseline | 3.74m | 15.3m | 20.2m | **30.1m** (killed) |
| + fast disk | 3.42m | 6.7m | 11.4m | 14.6m |
| + no `--with-deps` | **0.36m** | **7.0m** | **8.4m** | **12.7m** |

The harness step measured against every baseline run from the previous three days, whose means ranged 12.1–15.3m and worst legs 13.7–24.7m. Final legs: 4.5 5.2 6.6 6.7 6.9 7.3 7.5 11.2. Every one beats the previous best.

Headroom under the 30m cap went from 1.0x to 2.4x, so `timeout-minutes` needs no change and the shard reweight is no longer urgent.

Ubuntu is unaffected: the playwright step read 0.45 / 0.43 / 0.42 / 0.46m across the four runs.

## What I tried that did not work

Both are in the history as reverts, with the numbers that killed them:

- **Defender exclusions** (4b1b19f7, reverted in 4a8e7778). I was confident antivirus was scanning every unpacked file. Exclusions applied cleanly and changed nothing — 14.0m mean against a 12.1–15.3m baseline band, and the byte-identical 58 MB cache-restore probe did not move either.
- **Playwright browsers on `D:`** (c3f99b39, reverted in c9b8f054). Measured flat, 3.42m → 3.34m. The variable did apply; there was just little to win, because browser archives are a few large files and bulk throughput was never the bottleneck.

Ruled out earlier by measurement: network throughput. `actions/cache` pulled at 73.7 MB/s on Windows against 81.8 on Ubuntu — parity — while extracting the same 58 MB took 5.6s against 0.76s. That is what pointed at per-file cost rather than bandwidth.

The temporary benchmark used for cause 1 is removed again in 1770eb39.

## Not covered here

- Windows is still ~7x slower than Linux at file creates even on `D:` (4295–6008 vs 27539–32663 files/s). That is NTFS, not fixable in CI config.
- `PUPPETEER_CACHE_DIR` writes browser binaries into the git checkout and `.puppeteer-cache` is not in `.gitignore`. Pre-existing, left alone.
- Two unrelated red runs turned up while measuring, both since verified as one-off flakes that have not recurred in the last 12 runs:
  - Run [33662504119](https://github.com/apify/actor-templates/actions/runs/33662504119), `Test Python templates (windows-latest, 3.11, 2/2)` — the `Install Chrome and Chromedriver` step died on `Error: read ECONNRESET` fetching chromedriver. A dropped TLS connection, so `@puppeteer/browsers` printed its usage text and exited 1. Every other red job in that run shows `cancelled` because I cancelled the run to read its logs.
  - Run [33575265665](https://github.com/apify/actor-templates/actions/runs/33575265665) on master, `ts-crawlee-playwright-camoufox` exiting 1 from `apify run` with no crawler output. The next scheduled master run, [33699476711](https://github.com/apify/actor-templates/actions/runs/33699476711), is green on the same tree. No root cause established — worth a real look if it returns.
- Since these Windows fixes landed, the longest job in a run is an **Ubuntu Python** leg at 14m24s, driven by `python-playwright` crawling 56 live pages serially. Filed separately as #909; nothing in this PR affects it.


l2ysho:
@B4nan this may be interesting also for tooling

.github/workflows/test_llm_ai_node_templates.yaml
```diff
@@ -20,6 +20,21 @@ jobs:
     timeout-minutes: 30
 
     steps:
+      # The Windows runners have two disks: C: is the network-backed Azure OS disk,
```
DaveHanns, line 23: Are these comments necessary? We can, at least, shorten them a bit.

DaveHanns (approved):
Approved, but please review the long comments and try to come up with more concise and to-the-point formulation.

PR #915, opened 2026-09-04
chore: shorten the Windows temp-disk comment

l2ysho:
Follow-up to #907.

De-slop comments.


B4nan (approved):
that's a `chore`, not a `ci` related work really. also next time do it right in the PR instead of a follow up, even better, add a global memory to your claude.md to keep comments short/consise in general, preferrably one liners

(also such things dont need 3 reviewers, heh)

PR #917, opened 2026-09-04
fix(templates): port python-mcp-proxy to mcp 2 and bump fastmcp to v4

l2ysho:
> [!NOTE]
> **TL;DR**
> fastmcp 4  #911 could not be closed alone. All required stuff updated in this PR.

## Why #911 is red

[#911](https://github.com/apify/actor-templates/pull/911) bumps `fastmcp` to `>=4.0.2` but leaves `mcp >= 1.25.0, < 2.0.0` next to it. Every Python lint and type-check leg dies in ~8s, before linting starts:

```
× No solution found when resolving dependencies for split
╰─▶ Because fastmcp-slim[client]>=4.0.2 depends on mcp>=2.0.0,<3.0.0 and
    fastmcp>=4.0.2 depends on fastmcp-slim[client]==4.0.2 ... your project's
    requirements are unsatisfiable.
```

`renovate/artifacts` on that PR fails for the same reason — Renovate could not regenerate `uv.lock`.

## Why Renovate cannot finish it

Widening `mcp` to `>= 2.0.0, < 3.0.0` resolves, but `ty check` then reports **34 diagnostics, all in `templates/python-mcp-proxy/my_actor/mcp_gateway.py`**. mcp 2 is a breaking rewrite of the low-level server API the proxy is built on:

| Was (mcp 1) | Is (mcp 2) |
|---|---|
| `app.request_handlers[types.XRequest] = fn` | `Server(..., on_x=fn)` — constructor kwargs |
| handler takes a `Request`, returns `types.ServerResult(result)` | handler takes `(ctx, params)`, returns the result type directly |
| `types.ServerResult(...)` callable wrapper | plain union — not callable |
| `response.serverInfo` | `response.server_info` |
| `CallToolResult(isError=True)` | `CallToolResult(is_error=True)` |
| `req.params.progressToken` | `params.progress_token` |

And one break the type checker never saw, because the import is resolved at runtime:

```
ImportError: cannot import name 'streamablehttp_client' from 'mcp.client.streamable_http'
```

mcp 2 renamed it to `streamable_http_client`, dropped the per-call HTTP kwargs in favour of a pre-built client, and yields two streams instead of three.

## What this PR does

**Pins** — `pyproject.toml`, `templates/python-mcp-empty/requirements.txt`, `templates/python-mcp-proxy/requirements.txt`:

- `fastmcp` → `>= 4.0.2, < 5.0.0`
- `mcp` → `>= 2.0.0, < 3.0.0`

**Port** — `templates/python-mcp-proxy/my_actor/`:

- `mcp_gateway.py` rewritten against the `on_*` constructor API. Handlers are collected per advertised capability and passed to `Server(...)` in one go. Whitelist filtering, tool-call charging and the unauthorized-tool error path are unchanged.
- `server.py` — `streamable_http_client`, with headers/timeout/auth moved into a `create_mcp_http_client(...)` passed as `http_client=`.
- `models.py`, `server.py` — `httpx` → `httpx2`. mcp 2 runs on httpx2 and the `auth` field is handed straight to it, so the template would otherwise ship two HTTP stacks. `httpx` stays in the root env; `python-beautifulsoup` and `python-start` still use it.

## Behavior changes worth a reviewer's eye

Two handlers are **dropped**, because mcp 2 emits a runtime `MCPDeprecationWarning` at `Server(...)` construction for each:

- **`logging/setLevel` forwarding** — "The logging capability is deprecated as of 2026-07-28 (SEP-2577)."
- **client-to-server progress forwarding** — "Client-to-server progress is deprecated as of 2026-07-28; progress is server-to-client only."

`resources/subscribe` and `resources/unsubscribe` are **kept**, so the proxy still works in front of a 2025-era remote. Only the client-side calls are deprecated, not the handler slots, so they cost one `# ty: ignore[deprecated]` each and no runtime warning.

Say the word if you would rather keep logging and progress and accept the startup warnings.

## Verification

Local, against the real packages (fastmcp 4.0.2, mcp 2.1.1):

- `uv run poe lint` — clean. `uv run poe type-check` — `All checks passed!` (was 34 diagnostics).
- **Gateway end to end, in-memory**: upstream server → `create_gateway` → client. Tool list filtered to the whitelist, allowed call proxied, unauthorized call returns `is_error=True` with the right message, charging fired once with the configured count `('search_papers', 2)`, prompts proxied, upstream name/version forwarded.
- **HTTP branch end to end, over a real socket**: a FastMCP streamable-http server on one port, `ProxyServer(server_type=HTTP)` on another. Through the proxy: `initialize` reports the upstream's `upstream-http 9.9.9`, `list_tools` → `['echo']`, `call_tool` → `echo: hello`.
- **`python-mcp-empty` under fastmcp 4**: server builds, `list_tools` → `['add']`, `add(2,3)` returns the expected structured content, `http_app(transport='streamable-http')` still constructs.
- Both templates' `requirements.txt` resolve on their own (`arxiv-mcp-server` included).

## Follow-ups, not in this PR

- `SessionTrackingMiddleware` / `create_starlette_app` still call `session_manager.handle_request`, which mcp 2 keeps. No change needed, but the newer `asgi_app` is the forward path.
- `python-mcp-proxy/README.md:128` points at `src/mcp_gateway.py`; the file lives in `my_actor/`. Pre-existing.
- mcp 2 offers `on_subscriptions_listen` as the replacement for resource subscriptions. Wiring the proxy through it is a design change, not a port.

Close #911 once this is green.


l2ysho:
@vdusek can you pls take a look?

templates/python-mcp-empty/requirements.txt
```diff
@@ -3,6 +3,6 @@
 
 apify >= 4.0.0, < 5.0.0
 apify-client
-fastmcp >= 3.0.0, < 4.0.0
-mcp >= 1.25.0, < 2.0.0
+fastmcp >= 4.0.2, < 5.0.0
```
vdusek: Do 4.0.1 and 4.0.2 contain any critical bug fixes that our template depends on, meaning it wouldn't work with 4.0.0? If not, we should use as wide a version range as possible, probably `>= 4.0.0` and `< 5.0.0`.

This applies to other occurrences as well.

pyproject.toml
```diff
@@ -12,15 +12,16 @@ dependencies = [
     # crewai 1.x conflicts with the aggregate env (requires uv < 0.10, pydantic-settings < 2.11)
     # "crewai[tools] >= 1.12.2, < 2.0.0",
     "fastapi >= 0.135.0, < 1.0.0",
-    "fastmcp >= 3.0.0, < 4.0.0",
+    "fastmcp >= 4.0.2, < 5.0.0",
```
vdusek: same as https://github.com/apify/actor-templates/pull/917/changes#r3947918694

templates/python-mcp-proxy/requirements.txt
```diff
@@ -5,8 +5,8 @@ apify >= 4.0.0, < 5.0.0
 apify-client
 arxiv-mcp-server >= 0.3.1, < 1.0.0
 fastapi >= 0.135.0, < 1.0.0
-httpx >= 0.28.0, < 1.0.0
-mcp >= 1.25.0, < 2.0.0
+httpx2 >= 2.9.0, < 3.0.0
```
vdusek: same as for `fastmcp`, we should use `>= 2.0.0, < 3.0.0`, if possible

pyproject.toml
```diff
@@ -12,15 +12,16 @@ dependencies = [
...
     "httpx >= 0.28.0, < 1.0.0",
+    "httpx2 >= 2.9.0, < 3.0.0",
```
vdusek: same as for `fastmcp`, we should use `>= 2.0.0, < 3.0.0`, if possible

vdusek (approved):
Approving, but please consider my comments regarding the version constraints.

Also, we should probably migrate everywhere from `httpx` to `httpx2`, but that's for another PR.

vdusek (approved):
LGTM

PR #923, opened 2026-09-07
chore(templates): migrate python-start and python-beautifulsoup to httpx2

### LINKED ISSUE #922: Migrate python-start and python-beautifulsoup from httpx to httpx2
> [!NOTE]
> **TL;DR**
> Two Python templates still use `httpx`, while `python-mcp-proxy` now uses `httpx2`. Lets sync all to httpx2

## What is going on

`httpx2` is the next generation of `httpx` ([pydantic/httpx2](https://github.com/pydantic/httpx2)).

#917 moved `python-mcp-proxy` to `httpx2`, because `mcp` 2 requires it. The other two templates were left alone. So the repo now ships both clients, and the root `pyproject.toml` lists both.

## Why bother

- One HTTP client to document, teach and keep updated instead of two.
- `httpx` 0.x is the old line. New templates should start on the new one.
- Drops a duplicate dependency from the aggregate env.

## What to change

**`templates/python-start`**
- `requirements.txt:6` - `httpx >= 0.28.0, < 1.0.0` -> `httpx2 >= 2.5.0, < 3.0.0`
- `my_actor/main.py:20` - `from httpx import AsyncClient` -> `from httpx2 import AsyncClient`
- `README.md:5,15,21` - the HTTPX name and link

**`templates/python-beautifulsoup`**
- `requirements.txt:6` - same pin change
- `my_actor/main.py:15` - same import change
- `README.md:5,13` - the HTTPX name and link

**Root**
- `pyproject.toml` - drop `httpx >= 0.28.0, < 1.0.0`, keep `httpx2`
- `uv lock`

## Is it a drop-in?

For what these two templates do, yes. Both only use:

- `async with AsyncClient() as client`
- `await client.get(url, follow_redirects=True)`
- `response.text` and `response.raise_for_status()`

All of it exists in `httpx2` under the same names, with the same `follow_redirects=False` default. Checked against `httpx2` 2.12.0. No code logic needs to change - only the import line.

## Done when

- [ ] No `httpx` left in any template or in `pyproject.toml`
- [ ] `uv run poe lint` and `uv run poe type-check` pass
- [ ] Both templates still fetch a page and write rows to the dataset

## Where this came from

@vdusek's review on #917:

> Also, we should probably migrate everywhere from `httpx` to `httpx2`, but that's for another PR.

l2ysho:
> [!NOTE]
> **TL;DR**
> `python-start` and `python-beautifulsoup` now use `httpx2` instead of `httpx`, and the direct `httpx` pin is gone from the root env. The repo ships one HTTP client again.

This was stacked on #917 while that was open. #917 has since merged, and this branch is now rebased onto `master` — the diff below is this change alone.

## What changed

**`templates/python-start`** and **`templates/python-beautifulsoup`**:

- `requirements.txt` — `httpx >= 0.28.0, < 1.0.0` → `httpx2 >= 2.5.0, < 3.0.0` (same pin `python-mcp-proxy` already uses)
- `my_actor/main.py` — `from httpx import AsyncClient` → `from httpx2 import AsyncClient`, plus the comments naming the library
- `README.md` — HTTPX → HTTPX2, link to <https://github.com/pydantic/httpx2> (the package's declared Homepage; httpx2 has no separate docs site)

**Root**:

- `pyproject.toml` — direct `httpx` pin dropped; `httpx2` was already there from #917
- `uv.lock` — regenerated. Two lines removed. `httpx` stays in the lock as a transitive dep of the apify/crawlee/fastmcp line, which is expected.

No code logic changed. `AsyncClient()`, `client.get(url, follow_redirects=True)`, `response.content` / `.text` and `raise_for_status()` all exist in `httpx2` under the same names with the same defaults.

## How it was verified

| Check | Result |
|---|---|
| `uv run poe lint` | `78 files already formatted` / `All checks passed!` |
| `uv run poe type-check` | `All checks passed!` |
| `uv lock --check` | 288 packages, consistent |

Both templates run end to end, twice each — once against the aggregate repo env, once against a **fresh standalone venv built from the template's own `requirements.txt`** (which is what CI's `test-python-templates` does):

- **`python-start`**, input `{"url": "https://apify.com/"}` — exit 0, 36 headings pushed to the dataset, no errors.
- **`python-beautifulsoup`**, `max_depth: 0` — exit 0, 1 row with title + `h1s`/`h2s`/`h3s`. Also run at `max_depth: 1` — exit 0, 89 rows.

Standalone resolution installs `apify 4.0.2`, `beautifulsoup4 4.15.0`, `httpx2 2.12.0` and **no `httpx` at all** — so the template env is genuinely down to one HTTP stack, not just the direct pin.

One caught-and-logged fetch failure during the `max_depth: 1` crawl (`github.com/apify/crawlee`, a read timeout from that host). Handled by the template's existing `except`, run still exit 0. Not related to this change.

## Deliberately left out

- **`templates/manifest.json:53,92`** still list `"httpx"` in `technologies` for both templates. That array drives the template icon, which apify-web serves from `public/img/template-icons` — that directory has `httpx.svg` and **no `httpx2.svg`**. Flipping the value here would point both templates at a missing icon. It needs a sequenced cross-repo change: land `httpx2.svg` in apify-web first, then flip both entries. Happy to open a follow-up issue.
- `python-mcp-proxy/my_actor/models.py:31` has a comment reading `# Optional HTTPX authentication handler` next to an `httpx2.Auth` type. That line belongs to #917, so it is left for that PR.

Closes #922


vdusek (approved):
LGTM

PR #872, opened 2026-07-24
chore: Update templates for base image: apify/actor-node-playwright, apify/actor-node-playwright-chrome, apify/actor-node-playwright-firefox, apify/actor-node-playwright-webkit, apify/actor-node-playwright-camoufox

apify-service-account:
Automated update of templates for base image `apify/actor-node-playwright, apify/actor-node-playwright-chrome, apify/actor-node-playwright-firefox, apify/actor-node-playwright-webkit, apify/actor-node-playwright-camoufox`

**Parameters:**
- Base Image: `apify/actor-node-playwright, apify/actor-node-playwright-chrome, apify/actor-node-playwright-firefox, apify/actor-node-playwright-webkit, apify/actor-node-playwright-camoufox`
- Runtime Version: `24`
- Module Version: `1.61.1`

> Generated by [Update Templates](https://github.com/apify/actor-templates/actions/workflows/update-templates.yml) workflow.

PR #926, opened 2026-09-08
fix(ci): let update-templates refresh the lockfile it invalidates

l2ysho:
> [!NOTE]
> **TL;DR** This should be a leftover after npm -> pnpm migration.

## The failure

Every dispatch that changes the pinned Playwright or Puppeteer version dies:

```
[ERR_PNPM_OUTDATED_LOCKFILE] Cannot install with "frozen-lockfile" ...
  - playwright (lockfile: 1.61.1, manifest: 1.63.0)
```

Latest: [run 34120119724](https://github.com/apify/actor-templates/actions/runs/34120119724).

Master itself is green. This is the `Update Templates` workflow only, so it has been easy to miss.

## Origin

`scripts/actions/update-templates.mts` rewrites the module version in `package.json`, then runs `pnpm install` so `pnpm-lock.yaml` is back in sync before the bot commits both files.

That install is load-bearing. It was added in #734 precisely because #614 had started editing the root `package.json`, leaving every bot PR with a lockfile that disagreed with its own manifest. The PR this workflow opens installs with `--frozen-lockfile`, so a stale lockfile fails every check on it. Deleting the line would relocate the failure, not remove it.

07f5cff9 (#765, pnpm migration) changed that call:

```diff
-execSync('npm install', { cwd: new URL('../../', import.meta.url) });
+execSync('pnpm install', { cwd: new URL('../../', import.meta.url) });
```

`npm install` closes a manifest/lockfile gap silently. `pnpm install` defaults to `--frozen-lockfile` on CI and aborts, because a gap there usually means someone forgot to commit their lockfile. Here the gap is deliberate and one line old.

First failure came 3.5 hours after that merge.

## Impact

Last bot commit to bump the root pin: 458bbfe4, **2026-05-15**. Nothing since — roughly 3.7 months.

Current drift on master:

```
apify/actor-node-playwright:24-1.61.1            (playwright is at 1.63.0)
apify/actor-node-playwright-chrome:24-1.61.1
apify/actor-node-playwright-camoufox:24-1.59.1   (four versions behind)
apify/actor-node-puppeteer-chrome:24-24.43.1
```

Less bad than the gap suggests, because Renovate has been catching some of it as a side effect of root dependency updates. Renovate does not know about Apify base image tags, though, so they only move when they happen to ride along. The camoufox gap is what that looks like when they don't.

Not every Node dispatch died — #872 got through on 2026-07-24 because Renovate had already set the root pin to the target version nine days earlier, so the script wrote the same value and the lockfile stayed valid. The failure hits whenever the target differs from what the root `package.json` already holds, which is the normal case.

## The fix

```diff
-execSync('pnpm install', { cwd: new URL('../../', import.meta.url) });
+execSync('pnpm install --no-frozen-lockfile --lockfile-only', {
+    cwd: new URL('../../', import.meta.url),
+});
```

`--no-frozen-lockfile` restores the `npm install` behaviour the script was written against.

`--lockfile-only` because the workflow commits the lockfile and never runs the installed code, so downloading Playwright's browser binaries is wasted time.

## Verification

Ran the script end to end on a clean checkout with Node 24 and pnpm 11.13.0, using the workflow's own environment. Reproduced the original failure first, then confirmed the fix produces the same file set as the last successful bot commit (a0936b2d):

```
package.json                                          1.61.1 -> 1.63.0
pnpm-lock.yaml                                        specifier + version -> 1.63.0
templates/js-crawlee-playwright-camoufox/{Dockerfile,package.json}
templates/js-crawlee-playwright-chrome/{Dockerfile,package.json}
templates/ts-crawlee-playwright-camoufox/{Dockerfile,package.json}
templates/ts-crawlee-playwright-chrome/{Dockerfile,package.json}
templates/ts-playwright-test-runner/{Dockerfile,package.json}
```

No `node_modules` written, confirming `--lockfile-only`. Prettier passes.

## Deliberately not fixed

`pnpm-workspace.yaml` sets `minimumReleaseAge: 1440`, so a dispatch arriving within 24 hours of the npm publish still fails to resolve — a different error, `ERR_PNPM_NO_MATURE_MATCHING_VERSION`. That happened on [run 33933693188](https://github.com/apify/actor-templates/actions/runs/33933693188), which fired 2 hours after the 1.63.0 publish.

Leaving the 24-hour supply-chain protection alone was a deliberate call. The cost is small: dispatches repeat for the same image over several days, so only the first attempt after a release is lost. Playwright 1.63.0 was dispatched four times; three of those were outside the window and would have passed with this fix.

The alternatives, if anyone disagrees — add `playwright`, `playwright-core`, `@playwright/test`, `puppeteer` to `minimumReleaseAgeExclude`, or pass `--config.minimumReleaseAge=0` on this one install. Both verified working, both weaken the protection.

## Follow-ups, not in this PR

- The backlog does not clear itself — the workflow only fires on new base image publishes. Someone should trigger it manually via `workflow_dispatch` for playwright and camoufox once this lands.
- #872 has been open since 2026-07-24 with auto-merge armed, `BEHIND`, and red checks that ran against a much older master. Likely just needs a rebase and rerun; not confirmed.


l2ysho:
@B4nan can you take a look pls?

B4nan (approved):

PR #928, opened 2026-09-09
chore: Update Camoufox version in Python template 

Pijukatel:
Bump lower bound to make the Camoufox template compatible with the latest Playwright 1.61.0 and 1.62.0


vdusek (approved):
LGTM

PR #937, opened 2026-09-09
fix(templates): bound the crawl in the hand-rolled Python scrapers

l2ysho:
> [!NOTE] 
> **TL;DR**
> `python-playwright`, `python-selenium` and `python-beautifulsoup` crawl every link on the live apify.com homepage — 90+ pages — with no request cap, so a slow CI runner blows the 30-minute job budget. Cap them at 10 requests, the way the four Python Crawlee templates already do, and give selenium the per-page timeout the other two get for free. Also fixes a request-queue bug found while reviewing the change.

---

## The flake

[Run 34371908072](https://github.com/apify/actor-templates/actions/runs/34371908072) lost two Windows legs to the 30-minute job timeout. The re-run of the same commit was green in 11 minutes — same code, different outcome, so a flake rather than a regression.

These three templates drive a hand-rolled request-queue loop. The smoke run enqueues every `http(s)` link found on `https://apify.com` — 90-92 pages, about a third of them off-site. Run time is therefore:

```
(number of pages) × (seconds per page)
```

Neither factor was bounded. In the failing leg, 30 of python-playwright's 92 pages hit a 30 s navigation timeout — 15 minutes of dead wait — and the job died with python-selenium still mid-crawl. No test result was reported, and the shard's other three templates never ran.

Across 13 recent runs, 5 of 104 shard-1 Python legs blew up (31m, 31m, 30m, 30m, 26m) against a 6-8 minute median, with a long tail at 20m/18m/14m. Ubuntu was hit too, so this is not a Windows quirk.

## What changed

**Cap the page count (3 templates).** A `MAX_REQUESTS_PER_CRAWL = 10` constant and a counter in the loop condition. This is the convention the four `python-crawlee-*` templates already carry (`max_requests_per_crawl=10`). These three are the only crawling templates in the repo with **neither** a request cap **nor** a host restriction — the JS/TS browser templates without a cap scope their `enqueueLinks` to `globs: ['https://apify.com/*']` with a terminal handler.

**Bound the cost per page (python-selenium).** Capping the count is only half of it. Playwright defaults to a 30 s navigation timeout and httpx to 5 s, so the other two templates were already covered — selenium set nothing and fell through to the 120 s HTTP timeout between the client and chromedriver. `driver.set_page_load_timeout(30)` brings it in line. A slow page now raises `TimeoutException`, which the loop's existing handler logs before moving on, exactly as Playwright's timeout already does.

**Fix a request-queue bug (python-beautifulsoup).** Found reviewing the first commit. The `finally` block marked `new_request` — the last link enqueued, never fetched — instead of `request`. Crawlee short-circuits on a request that is not in progress, so no request was ever marked handled. A/B on a local run: before, 10 `Marking request ... as handled that is not in progress` warnings and `handled_request_count: 0`; after, no warnings and `handled_request_count: 10`.

## Verification

Local, same machine, all three templates:

| | pages crawled | wall time |
|---|---|---|
| before | 272 | 234.7 s |
| after | 30 (10 each) | 49.3 / 48.7 / 45.4 s |

Three consecutive runs gave an identical page count, so it is deterministic now rather than sampled from apify.com's link graph.

CI on the first push of this branch ([run 34389302327](https://github.com/apify/actor-templates/actions/runs/34389302327)): all 16 Python jobs green, worst leg 12 minutes against a 31-minute cancel before. That run is also what exposed the selenium gap — with the crawl already capped, python-selenium still ranged 30 s to 441 s across the eight shard-1 legs, and two single pages in the slowest leg took 111 s and 120 s. Hence the second commit.

Selenium's page-load timeout was verified against a blackholed address: raises `TimeoutException` at the limit, driver stays usable afterwards. Two local runs of the template: 10 pages, 20.5 s and 20.7 s.

`eslint`, `prettier`, `ruff check`, `ruff format --check`, `ty check`, and the non-template Jest suite (17/17) all pass.

## Reverted along the way

An earlier commit added an 8-minute `spawnSync` ceiling to `checkTemplateRun`, so a hung run could not consume the whole job silently. The number was wrong: `ts-crawlee-playwright-camoufox` legitimately runs up to **10m12s** ([run 34330088221](https://github.com/apify/actor-templates/actions/runs/34330088221)), and the ceiling failed a healthy run on this PR. Reverted in 53e19fd0; the harness is unchanged from master.

The idea still has merit — today a hung run kills the job with no test result at all, which is why the original failure took so long to diagnose. It needs a ceiling derived from a sweep of every template's run duration, so it belongs in its own PR.

## For the reviewer

- **Behavior change for users.** Anyone copying these three templates now gets a 10-page demo, and selenium gives up on a page after 30 s. Both match what the rest of the catalog already does.
- **I could not reproduce the original CI failure locally** — my network hit zero page timeouts, so the 30 s multiplier never fired. What is verified locally is the multiplicand: the page count. CI covers the rest.
- **Known remaining soft spot, not addressed here:** the camoufox templates. The Node job in the run above logged `Memory is critically overloaded. Using 3820 MB of 3997 MB (96%)` with 26 navigation timeouts, and `ts-crawlee-playwright-camoufox`'s whole test ranges 81 s to 663 s. Different templates, different cause.
- **One low-severity finding left unfixed**, outside this diff: `page` can be unbound in python-playwright's `finally` if `context.new_page()` raises.
- **`dist/` archives are untouched** — `build_archives.yaml` regenerates them on merge to master.

### DIFF
--- a/templates/python-beautifulsoup/my_actor/main.py
+++ b/templates/python-beautifulsoup/my_actor/main.py
@@ -14,7 +14,7 @@ from apify import Actor, Request
 from bs4 import BeautifulSoup
 from httpx import AsyncClient
 
-# Limit the crawl to max requests. Remove or increase it for crawling all links.
+# Limit the crawl to max requests. Increase it to crawl more links.
 MAX_REQUESTS_PER_CRAWL = 10
 
 
@@ -102,5 +102,5 @@ async def main() -> None:
 
                 finally:
                     # Mark the request as handled to ensure it is not processed again.
-                    await request_queue.mark_request_as_handled(new_request)
+                    await request_queue.mark_request_as_handled(request)
                     handled_requests += 1
--- a/templates/python-beautifulsoup/my_actor/main.py
+++ b/templates/python-beautifulsoup/my_actor/main.py
@@ -14,6 +14,9 @@ from apify import Actor, Request
 from bs4 import BeautifulSoup
 from httpx import AsyncClient
 
+# Limit the crawl to max requests. Remove or increase it for crawling all links.
+MAX_REQUESTS_PER_CRAWL = 10
+
 
 async def main() -> None:
     """Define a main entry point for the Apify Actor.
@@ -46,8 +49,10 @@ async def main() -> None:
 
         # Create an HTTPX client to fetch the HTML content of the URLs.
         async with AsyncClient() as client:
+            handled_requests = 0
+
             # Process the URLs from the request queue.
-            while request := await request_queue.fetch_next_request():
+            while handled_requests < MAX_REQUESTS_PER_CRAWL and (request := await request_queue.fetch_next_request()):
                 url = request.url
 
                 if not isinstance(request.user_data['depth'], (str, int)):
@@ -98,3 +103,4 @@ async def main() -> None:
                 finally:
                     # Mark the request as handled to ensure it is not processed again.
                     await request_queue.mark_request_as_handled(new_request)
+                    handled_requests += 1
--- a/templates/python-playwright/my_actor/main.py
+++ b/templates/python-playwright/my_actor/main.py
@@ -18,7 +18,7 @@ from playwright.async_api import async_playwright
 # When running on the Apify platform, these dependencies are already included
 # in the Actor's Docker image.
 
-# Limit the crawl to max requests. Remove or increase it for crawling all links.
+# Limit the crawl to max requests. Increase it to crawl more links.
 MAX_REQUESTS_PER_CRAWL = 10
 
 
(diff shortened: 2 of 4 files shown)


vdusek (approved):
LGTM, just one question: did you add the requests-per-crawl limit to all templates where we do crawling?

l2ysho:
> LGTM, just one question: did you add the requests-per-crawl limit to all templates where we do crawling?

These 4 have `globs: ['https://apify.com/*']` `label: 'detail',`, no flakyness for now so I did not touch it.

js-crawlee-playwright-chrome
js-crawlee-puppeteer-chrome
js-crawlee-playwright-camoufox
ts-crawlee-puppeteer-chrome

PR #946, opened 2026-09-11
chore: Update templates for base image: apify/actor-node-playwright-camoufox

apify-service-account:
Automated update of templates for base image `apify/actor-node-playwright-camoufox`

**Parameters:**
- Base Image: `apify/actor-node-playwright-camoufox`
- Runtime Version: `24`
- Module Version: `1.60.0`

> Generated by [Update Templates](https://github.com/apify/actor-templates/actions/workflows/update-templates.yml) workflow.


l2ysho:
Docker build fails on the playwright version check (Dockerfile pin vs package.json 1.63.0), plus a transient camoufox GitHub releases fetch timeout — fix the pin and re-run.

